PR #6, opened 2017-11-15
Drop clang/16.04 from build matrix.

RAOF:
We're only using clang to catch additional errors/warnings over gcc, so the fact that
it throws an error on 16.04's gtest is not interesting to us.


RAOF:
Superseded by the lxd-on-Travis branch.

PR #7, opened 2017-11-15
Fix FTBFS on Fedora rawhide

AlanGriffiths:


RAOF:
Yup, I was expecting we'd need to resync the FindGtestGmock with Mir's one...

bors r+

Saviq:
bors r=RAOF

PR #9, opened 2017-12-08
Add first input tests

RAOF:
This adds
- Fake pointer infrastructure, for sending input events through the compositor
- A mechanism for positioning windows, so that they are in a known location
- The Mir integration changes necessary to implement the above for Mir.


RAOF:
Looks like transient network error.

bors try

PR #10, opened 2017-12-11
Add a test for when a client lies about SHM buffer stride.

RAOF:
In this case everything can look ok, but accessing the pixels with the correct
pixel format will fail.

This is, among other things, a regression test for Mir issue https://github.com/MirServer/mir/issues/78


RAOF:
Looks like transient network failure (again?)

bors try

PR #12, opened 2018-01-12
More surface input tests

RAOF:
Fill out the other “pointer crosses a surface corner/edge” cases,
and add a “surface is moved under the cursor” test.


RAOF:
bors try

include/in_process_server.h
```diff
@@ -36,6 +36,7 @@ class Pointer
     Pointer(Pointer&&);
 
     void move_to(int x, int y);
+    void move(int dx, int dy);
```
AlanGriffiths: It would be consistent to call this "move_by"

PR #13, opened 2018-01-12
Use the LXD snap rather than a PPA

RAOF:
The ubuntu-lxc/lxd-stable PPA appears to have disappeared‽


RAOF:
Winner winner, chicken dinner!

PR #14, opened 2018-01-16
cmake: Copy Mir's GoogleTest machinery to fix Ninja build

RAOF:


RAOF:
bors try

RAOF:
Self-approving; this is trivial
bors r+

PR #15, opened 2018-01-16
bors: Clean up branches on merge

RAOF:


RAOF:
bors r+

Excited by the Ubuntu 17.04 image having a broken sources.list for a while!

PR #18, opened 2018-03-22
made wlcs::Client store shell surfaces in a vector

wmww:
made wlcs::Client store shell surfaces in a vector so they are all properly cleaned up


src/in_process_server.cpp
```diff
@@ -287,7 +288,8 @@ class wlcs::Client::Impl
     {
         Surface surface{client};
 
-        shell_surface = wl_shell_get_shell_surface(shell, surface);
+        wl_shell_surface * shell_surface = wl_shell_get_shell_surface(shell, surface);
```
RAOF: Coding style; the `*` should bind to the type name.

wmww: ah shoot

PR #28, opened 2018-05-03
self-test

### LINKED ISSUE #27: Don't add one listener to two different lists
After tracing through the libwayland code I found that when creating two clients things are corrupted because a listener gets added to two lists. (This meant I had to workaround things with the Copy, Cut & Paste tests.) 

From my debug output:

```
*************** wl_signal_add signal=0xd29d6c0
*************** wl_signal_add, l=0xd1ef938, l->notify=0x57757e
...
*************** wl_signal_add signal=0xd2abf00
*************** wl_signal_add, l=0xd1ef938, l->notify=0x57757e
```

AlanGriffiths:
Don't link the same wl_listener into every client. (Fixes #27)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,7 +73,7 @@ add_library(
   tests/test_bad_buffer.cpp
   tests/test_surface_events.cpp
   tests/frame_submission.cpp
-)
+        tests/self_test.cpp)
 
 target_link_libraries(
   wlcs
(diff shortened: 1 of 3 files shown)


AlanGriffiths:
```

+ apt-get update
E: Type 'curity' is not known on line 50 in source list /etc/apt/sources.list
E: The list of sources could not be read.
```
Seems unrelated and possibly test infrastructure. Let's see:

bors try

RAOF:
bors try

PR #29, opened 2018-05-04
Vendored in protocols

wmww:
Needed the XDG shell protocol in here to write tests for it. Added `src/protocol` and `src/generated` directories similar to how it is in `frontend_wayland`. (adding exactly 22,222 LOC was not deliberate)


wmww:
bors try

wmww:
@RAOF I'd like approval on the changes before merging.

PR #31, opened 2018-05-08
First xdg shell tests

wmww:


AlanGriffiths:
bors try

tests/test_xdg_surface_v6.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include <gmock/gmock.h>
+
+using XdgSurfaceV6Test = wlcs::InProcessServer;
+
+TEST_F(XdgSurfaceV6Test, supports_xdg_shell_v6_protocol)
+{
+    using namespace testing;
+
+    wlcs::Client client{the_server()};
+    ASSERT_TRUE(client.xdg_shell_v6() != nullptr);
```
RAOF: It's probably better to use `client.acquire_interface()` here? That throws on failure, so you don't need the `ASSERT_TRUE`

(And I intend to make it skip when the interface is *expected* not to exist)

include/xdg_shell_v6.h
```diff
@@ -0,0 +1,115 @@
...
+    void add_configure_notification(std::function<void(uint32_t)> notification)
+    {
+        configure_notifiers.push_back(notification);
+    }
+
+    operator zxdg_surface_v6*() const {return shell_surface;}
+
+    wlcs::Client* const client;
+    wlcs::Surface* const surface;
+    zxdg_surface_v6* shell_surface;
```
AlanGriffiths: Nit: This should be private as non-members should not update it. Consider a conversion operator for non-member access?

wmww: Done

wmww:
There you go. Missed that bit in original comment. I like those too.

tests/xdg_toplevel_v6.cpp
```diff
@@ -0,0 +1,208 @@
...
+        });
+
+    // default values
+    ASSERT_THAT(state, Ne(std::experimental::nullopt));
+    EXPECT_THAT(state.value().width, Eq(0));
+    EXPECT_THAT(state.value().height, Eq(0));
+    EXPECT_THAT(state.value().maximized, Eq(false));
+    EXPECT_THAT(state.value().fullscreen, Eq(false));
+    EXPECT_THAT(state.value().resizing, Eq(false));
+    EXPECT_THAT(state.value().activated, Eq(true));
```
RAOF: You could possibly make this more awesome with a Matcher for ::State.

Issue #30, opened 2018-05-04
BadBufferTest.test_truncated_shm_file gets Bus error


wmww:
Interestingly, this only happens when at least one instance of `source.roundtrip()` or `sink.roundtrip()` is present in `CopyCutPaste.given_source_has_offered_data_sink_sees_offer` (Don't ask me how long it took to figure that out). Leads me to believe those functions are installing a SIGBUS handler, which would be the manipulation of global state needed for one test to break another, and which is documented to break bad SHM buffer detection here: https://people.freedesktop.org/~whot/wayland-doxygen/wayland/Server/structwl__shm__buffer.html. Exactly where the SIGBUS handler is being installed I could not easily find, so I'll leave that to whoever wrote the code (probably @RAOF?)

My output from latest master WLCS and latest master Mir:
```
[ RUN      ] BadBufferTest.test_truncated_shm_file
[2018-05-04 10:48:49.515919] mirserver: Selected driver: mir:stub-graphics (version 0.31.2)
[2018-05-04 10:48:49.516267] <DEBUG> mirserver: No session management supported
[2018-05-04 10:48:49.516760] mirserver: Starting
[2018-05-04 10:48:49.522641] mirserver: Using software cursor
[2018-05-04 10:48:49.524128] mirserver: Selected input driver: mir:stub-input (version: 0.31.2)
[2018-05-04 10:48:49.524767] mirserver: Mir version 0.31.2
[2018-05-04 10:48:49.525031] mirserver: Initial display configuration:
[2018-05-04 10:48:49.525049] mirserver: * Output 1: VGA connected, used
[2018-05-04 10:48:49.525059] mirserver: . |_ Physical size 0.0" 0x0mm
[2018-05-04 10:48:49.525066] mirserver: . |_ Power is on
[2018-05-04 10:48:49.525072] mirserver: . |_ Current mode 1600x1600 60.00Hz
[2018-05-04 10:48:49.525078] mirserver: . |_ Preferred mode 1600x1600 60.00Hz
[2018-05-04 10:48:49.525082] mirserver: . |_ Orientation normal
[2018-05-04 10:48:49.525087] mirserver: . |_ Logical size 1600x1600
[2018-05-04 10:48:49.525093] mirserver: . |_ Logical position +0+0
[2018-05-04 10:48:49.525100] mirserver: . |_ Scaling factor: 1.00
Bus error (core dumped)
```

AlanGriffiths:
Possibly a dup of #27?

[...]

PR #34, opened 2018-05-14
Re-enable and fix copy/paste tests

wmww:
The copy/paste test had been disabled as a workaround for #30. The real problem is running *any* tests that send buffers before the bad buffer tests, so I simply moved the copy/paste one in CMake and re-enabled it.

commit bcdc705509e78bd6752cd4bd2af145f1e123115c broke the copy/paste test by causing `Client::create_visible_surface()` to block while the needed callbacks were called (thus they were called before the listener was installed). I resolved this issue by manually inlining some of `create_visible_surface()` into `CCnPClient`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,11 +69,12 @@ add_library(
   src/generated/xdg-shell-unstable-v6.c
   src/generated/xdg-shell.c
 
-  tests/copy_cut_paste.cpp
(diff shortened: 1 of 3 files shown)


AlanGriffiths:
Thanks for diagnosing this. It has just surfaced again on my list.

I'm still thinking about whether inlining `create_visible_surface()` is the best solution...

PR #35, opened 2018-05-15
Input region tests

wmww:
Tests surfaces with input regions with the same test style as `test_surface_events.cpp` uses

### DIFF
--- a/tests/surface_input_regions.cpp
+++ b/tests/surface_input_regions.cpp
@@ -227,7 +227,7 @@ InputRegion const multi_rect_region{"multi rect", {
      RegionAndMotion::window_width - small_rect_inset * 2, RegionAndMotion::window_height / 2}}};
 
 INSTANTIATE_TEST_CASE_P(
-    MultiRectRegion,
+    MultiRectRegionEdges,
     InputRegionPointerEnterTest,
     testing::Values(
         RegionAndMotion{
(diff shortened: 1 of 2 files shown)


tests/surface_input_regions.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+using namespace testing;
+
+// No tests use SUBTRACT_RECT as that is not yet implemented in Mir
+
+enum RegionAction
+{
+    ADD_RECT,
+    SUBTRACT_RECT
+};
```
RAOF: We'd normally use an `enum class` and CamelCase for the elements

tests/surface_input_regions.cpp
```diff
@@ -0,0 +1,255 @@
...
+    {
+        switch(e.action)
+        {
+        case ADD_RECT:
+            wl_region_add(wl_region, e.x, e.y, e.width, e.height);
+            break;
+        case SUBTRACT_RECT:
+            wl_region_subtract(wl_region, e.x, e.y, e.width, e.height);
+            break;
+        default:
```
RAOF: If you want to you can `#ifndef __clang__` here; clang will believe that the above is exhaustive

tests/surface_input_regions.cpp
AlanGriffiths: > We'd normally use an enum class and CamelCase for the elements

Actually our [guidelines](https://mir-server.io/doc/cppguide/index.html?showone=Enumerator_Names#Enumerator_Names) say lower_case for the members.

tests/surface_input_regions.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+using namespace testing;
+
+// No tests use Subtract as that is not yet implemented in Mir
+
+enum class RegionAction
+{
+    Add,
+    Subtract
+};
```
AlanGriffiths: Sorry, I didn't realize I was commenting on an obsolete review.

The enum class members [should be](https://mir-server.io/doc/cppguide/index.html?showone=Enumerator_Names#Enumerator_Names) lower_case.

PR #36, opened 2018-05-15
More copy, cut and paste testing

AlanGriffiths:


tests/copy_cut_paste.cpp
```diff
@@ -28,36 +27,32 @@
 using namespace testing;
 using namespace wlcs;
 
-using CopyCutPaste = wlcs::InProcessServer;
-
 namespace
 {
+struct StartedInProcessServer : InProcessServer
```
wmww: What is the need for this class? By keeping the server started between test cases, does it not risk one test corrupting another?

AlanGriffiths:
On Tuesday, 15 May 2018 18:16:17 BST, William W Wold 
<notifications@github.com> wrote:
> wmww commented on this pull request.
> 
> 
> 
>>  namespace
>  {
> +struct StartedInProcessServer : InProcessServer
> 
> What is the need for this class? By keeping the server started 
> between test cases, does it not risk one test corrupting 
> another?

It starts the server before each test fixture is created allowing the 
server to be used in constructing the fixture.

A fixture is constructed for each test, so there's always a new server.



-- 
Alan Griffiths. +44 (0)798 9938 758
Octopull Limited. http://www.octopull.co.uk/

PR #45, opened 2018-06-05
[travis] simplify lxd install

Saviq:


Saviq:
OK it seems like this breaks something badly. Can't touch this.

PR #47, opened 2018-06-08
Specify the Wayland socket to both server and client to avoid connecting to a Wayland desktop.

AlanGriffiths:


AlanGriffiths:
bors try

PR #52, opened 2018-06-19
added resources list to Client::Impl

wmww:


src/in_process_server.cpp
```diff
@@ -337,14 +337,14 @@ class wlcs::Client::Impl
...
-        xdg_surfaces_v6.clear();
-        xdg_toplevels_v6.clear();
         if (seat) wl_seat_destroy(seat);
         if (pointer) wl_pointer_destroy(pointer);
         if (touch) wl_touch_destroy(touch);
         if (data_device_manager) wl_data_device_manager_destroy(data_device_manager);
         if (xdg_shell_v6) zxdg_shell_v6_destroy(xdg_shell_v6);
         if (xdg_shell_stable) xdg_wm_base_destroy(xdg_shell_stable);
+        for (auto callback: destruction_callbacks) callback();
+        destruction_callbacks.clear();
```
RAOF: There's no particularly good reason to clear the callback list here, is there?

src/in_process_server.cpp
```diff
@@ -337,14 +337,14 @@ class wlcs::Client::Impl
...
         wl_shell_surfaces.clear();
-        xdg_surfaces_v6.clear();
-        xdg_toplevels_v6.clear();
         if (seat) wl_seat_destroy(seat);
         if (pointer) wl_pointer_destroy(pointer);
         if (touch) wl_touch_destroy(touch);
         if (data_device_manager) wl_data_device_manager_destroy(data_device_manager);
         if (xdg_shell_v6) zxdg_shell_v6_destroy(xdg_shell_v6);
         if (xdg_shell_stable) xdg_wm_base_destroy(xdg_shell_stable);
+        for (auto callback: destruction_callbacks) callback();
```
RAOF: I'd vaguely prefer this to be on two lines.

src/in_process_server.cpp
wmww: I want to drop all the callbacks before `wl_display_disconnect(display)`, in case they own things that need to be dropped. In practice they just own shared pointers that are reset, but this is a bit more robust to future uses.

PR #55, opened 2018-06-26
Improved input region tests

wmww:
This tests both XDG shell surfaces and WL shell surfaces. I split up the tests so that first every edge/corner of every input region we have is tested once. Also each setup and input variation (wl_shell and xdg_shell, pointer and touch, surface and subsurface) are tested against a subset of input region edges and corners that is likely to expose bugs. This way we can have virtually the same test coverage with only 90 tests, instead of the hundreds (and eventually thousands) it would be otherwise. Also maintenance is easier. gtest does not appear to support multi dimensional parameterized tests, so code size would grow exponentially if we don't organize this way.


AlanGriffiths:
Try again:

bors r+

PR #58, opened 2018-07-06
fix-master

AlanGriffiths:
There's a race condition in some tests, so fix that:

    XdgToplevelV6Configuration.maximized_and_unmaximized
    XdgToplevelV6Configuration.fullscreened_and_restored


wmww:
Closing, as this appears to conflict with the protocol: https://github.com/MirServer/mir/blob/8be8853f71adbc016b05692bcf6e2ebe30d04cb2/src/server/frontend_wayland/protocol/xdg-shell-unstable-v6.xml#L689

wmww:
Upon further investigation, that seems to only apply to the client. There is no mention of double bufferedness in the `zxdg_toplevel_v6_set_maximized` description, and both weston and wlroots send the configure event before the client commits.

PR #59, opened 2018-07-07
Better XDG shell tests

wmww:
This also adds the ability to click, which isn't currently used but is nice to have.

### DIFF
--- a/example/mir_integration.cpp
+++ b/example/mir_integration.cpp
@@ -741,6 +741,40 @@ void wlcs_pointer_move_absolute(WlcsPointer* pointer, wl_fixed_t x, wl_fixed_t y
     wlcs_pointer_move_relative(pointer, wl_fixed_from_double(rel_x), wl_fixed_from_double(rel_y));
 }
 
+void wlcs_pointer_button_down(WlcsPointer* pointer, int button)
+{
+    auto device = reinterpret_cast<FakePointer*>(pointer);
+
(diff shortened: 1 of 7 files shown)


example/mir_integration.cpp
```diff
@@ -741,6 +741,40 @@ void wlcs_pointer_move_absolute(WlcsPointer* pointer, wl_fixed_t x, wl_fixed_t y
...
+        std::chrono::steady_clock::now().time_since_epoch());
+
+    auto event_sent = device->runner->event_listener->expect_event_with_time(event_time);
+
+    device->pointer->emit_event(
+        mir::input::synthesis::a_button_down_event()
+            .of_button(button)
+            .with_event_time(event_time));
+
+    event_sent->wait_for(std::chrono::seconds{5});
```
AlanGriffiths: `std::chrono::seconds{5}` should probably be extracted as a constant. E.g.
```
event_sent->wait_for(max_send_delay);
```
Also, if sending fails we should be reporting it - as it will probably cause a test failure

wmww: Turns out the constant already existed, so switched to using that. Not sure exactly how I should be reporting timeout failures. Maybe throwing an exception to fail the test?

AlanGriffiths: Failing an EXPECT (or an ASSERT) probably works best. I don't think gmock has a way to output messages.

PR #60, opened 2018-07-13
Improve fake event sending code

wmww:
Abstracts code to send a fake Mir event to a function and shows a failure message if waiting for an event fails


wmww:
bors try

PR #61, opened 2018-07-19
Made subsurface tests parameterized

wmww:
Tests subsurfaces
* On WL shell surfaces with mouse input
* On XDG shell surfaces with mouse input
* On default shell surfaces with touch input

### DIFF
--- a/tests/subsurfaces.cpp
+++ b/tests/subsurfaces.cpp
@@ -26,16 +26,36 @@
 
 using namespace testing;
 
-class SubsurfaceTest : public wlcs::StartedInProcessServer
+struct SubsurfaceTestParams
+{
+    std::string name;
+    std::function<std::unique_ptr<wlcs::Surface>(wlcs::InProcessServer& server,
+                                                 wlcs::Client& client,
+                                                 int x, int y,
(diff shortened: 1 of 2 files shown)


tests/subsurfaces.cpp
```diff
@@ -26,16 +26,103 @@
...
+
+struct PointerInputDevice : AbstractInputDevice
+{
+    PointerInputDevice(wlcs::Server& server, wlcs::Client& client):
+        client{client},
+        pointer{server.create_pointer()}
+    {
+    }
+
+    void down_at(int x, int y) override
```
AlanGriffiths: The name doesn't seem to match the implementation. Which is right?

wmww: It's an override of an abstract method. 'down_at' doesn't make sense for pointers and 'move_to' doesn't make sense for touch. Maybe there's a name out there that's applicable to both, but I just choose one and went with it.

AlanGriffiths: place_at()?

wmww: Maybe `to_screen_position` is a bit better

PR #64, opened 2018-08-04
XDG popup tests

wmww:
Added XDG shell tests, including popup/positioner tests


wmww:
I've split this up into #66 and #67.

PR #66, opened 2018-08-06
XDG toplevel geometry offset tests

wmww:
* Added two tests for window geometry offset
* Moved `ConfigurationWindow`
* Suffixed names with `*Test`

This was part of #64, which I'm splitting up because it was a bit big and disorganized.


wmww:
For consistency, and to avoid name collisions (I don't want yet another `XdgToplevelV6` floating around)

PR #67, opened 2018-08-06
Added XDG popup and positioner tests

wmww:
This was part of #64, which I'm splitting up because it was a bit big and disorganized.

### DIFF
--- a/tests/xdg_popup_v6.cpp
+++ b/tests/xdg_popup_v6.cpp
@@ -1,0 +1,66 @@
+/*
+ * Copyright © 2012 Intel Corporation
+ * Copyright © 2013 Collabora, Ltd.
+ * Copyright © 2018 Canonical Ltd.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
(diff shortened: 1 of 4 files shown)


tests/xdg_popup_v6.cpp
```diff
@@ -0,0 +1,270 @@
...
+            });
+    }
+
+    std::experimental::optional<std::pair<int, int>> get_popup_position()
+    {
+        // state is not set, as WLCS window manager does not send the proper configure events.
+        // EXPECT_THAT(state.x, Eq(0));
+        // EXPECT_THAT(state.y, Eq(0));
+        // EXPECT_THAT(state.width, Eq(popup_width));
+        // EXPECT_THAT(state.height, Eq(popup_height));
```
AlanGriffiths: Is there a "WLCS window manager"? Do you mean the one use by miral-integration?

If so, this is a bug? Do we have an issue/pr?

wmww: Ah, that was written before the switch to a miral window manager. I'll see check if we're getting the right events now.

wmww: Resolved

PR #71, opened 2018-12-21
Switch to providing a wlcs binary test runner

RAOF:
This switches wlcs to ship a `wlcs` binary that will load a server integration DSO rather than a `libwlcs` that needs to be linked into a binary by the consuming project.

This, in turn, requires switching from the `__attribute__(weak)` approach to linking in integration symbols. Instead we load an `entry_point` vtable struct and acquire further vtable structs from there.


RAOF:
This, obviously, breaks the Mir tests and should not be pulled into Mir until I've done the extra work to make this autopush a package into our PPA and stuff like that.

https://github.com/MirServer/mir/pull/679 can be used to test (although you'll need to replace the hard-coded include path)

src/helpers.cpp
```diff
@@ -105,4 +108,18 @@ int wlcs::helpers::get_argc()
...
+void wlcs::helpers::set_module_under_test(char const* dso_name)
+{
+    ::dso_name = dso_name;
+}
+
+std::shared_ptr<WlcsServerIntegration const> wlcs::helpers::get_test_hooks()
+{
+    auto dso = std::make_shared<wlcs::SharedLibrary>(::dso_name);
+
+    auto entry_point = dso->load_function<WlcsServerIntegration const*>("entry_point");
```
AlanGriffiths: Shouldn't the entry point be named something more specific than "entry_point"?

At least let's namespace it with a "wlcs_" prefix.

src/helpers.cpp
```diff
@@ -105,4 +108,18 @@ int wlcs::helpers::get_argc()
 char const** wlcs::helpers::get_argv()
 {
     return ::argv;
-}
\ No newline at end of file
+}
+
+void wlcs::helpers::set_module_under_test(char const* dso_name)
+{
+    ::dso_name = dso_name;
```
AlanGriffiths: Here, or in the calling code, we should check for nullptr. (User input can never be trusted.)

RAOF:
> I'm happy with the general approach but: WlcsServerIntegration is unversioned and that will become a problem as soon as an additional interaction is needed. At the very least there should be some sort of version information that can be checked.

Hm, I think you must have missed the versioning that's done? Each of the structs has a `uint32_t version` as its first member, and we check the version we get in the various setup places.

This doesn't let a server-integration module provide multiple different versions depending on what the `wlcs` binary wants, but it does let `wlcs` handle (at the moment, by rejecting) different server-integration versions.

src/helpers.cpp
RAOF: Renamed as `wlcs_server_integration`

src/helpers.cpp
RAOF: I've moved much of this around; we now give actual error messages on invalid input.

include/shared_library.h
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    template<typename FunctionPtr>
+    FunctionPtr load_function(std::string const& function_name) const
+    {
+        return load_function<FunctionPtr>(function_name.c_str());
+    }
+
+    template<typename FunctionPtr>
+    FunctionPtr load_function(std::string const& function_name, std::string const& version) const
```
AlanGriffiths: I "just happen" to know the implementation of this doesn't compile on Alpine Linux. We don't use it (yet), so let's get rid of it.

include/in_process_server.h
```diff
@@ -37,7 +42,7 @@ class Pointer
 {
 public:
     ~Pointer();
-    Pointer(Pointer&&);
+      Pointer(Pointer&&);
```
AlanGriffiths: Stray whitespace?

.travis.yml
```diff
@@ -9,29 +8,81 @@ branches:
...
-    - snapd
+  snaps:
+  -  lxd
+
+stages:
+- name: test
+  # don't test on pushes to master or release pull requests or tag events
+  # TODO: enable release pull request testing when we have more CI power
+  if: NOT (( branch = master AND type = push )
+             OR ( head_branch =~ ^release/[\d\.]+$ ))
```
Saviq: I think we have enough power to keep this enabled drop the TODO.

.travis.yml
```diff
@@ -9,29 +8,81 @@ branches:
...
+            OR tag =~ ^v[\d\.]+$ )
 
 install:
-  - sudo snap install lxd
-  - sudo addgroup --system lxd
   - sudo adduser $USER lxd
-  - sudo snap start lxd
-  - for I in 1 2 3 4 5 6 7 8 9 10 ; do [ -S /var/snap/lxd/common/lxd/unix.socket ] || sleep 5 ; done
-  - sudo lxd init --auto
+  - sudo apt remove --assume-yes lxd
```
Saviq: Why this?

.travis.yml
```diff
@@ -9,29 +8,81 @@ branches:
...
+  - sudo /snap/bin/lxd init --auto
```
Saviq: I thought the snap does this on its own? Any case, `snap run lxd` is more idiomatic, if `lxd` itself isn't on `$PATH`.

debian/rules
```diff
@@ -0,0 +1,38 @@
...
+COMMON_CONFIGURE_OPTIONS =\
+  -DWLCS_BUILD_ASAN=ON \
+  -DWLCS_BUILD_UBSAN=ON
+
+$(info Building for $(DEB_DISTRIBUTION) on $(DEB_HOST_ARCH))
+
+ifeq ($(filter i386 armhf, $(DEB_HOST_ARCH)),)
+  ifeq ($(DEB_DISTRIBUTION),xenial)
+    ifneq ($(filter arm64 ppc64el, $(DEB_HOST_ARCH)),)
+      $(info Hello from xenial-arm64 or xenial-ppc64el)
```
Saviq: That left on purpose?

debian/rules
```diff
@@ -0,0 +1,38 @@
...
+    else
+      COMMON_CONFIGURE_OPTIONS += -DWLCS_BUILD_TSAN=ON
+    endif
+  else
+    COMMON_CONFIGURE_OPTIONS += -DWLCS_BUILD_TSAN=ON
+  endif
+else
+  # i386 and armhf do not have tsan
+  COMMON_CONFIGURE_OPTIONS += -DWLCS_BUILD_TSAN=OFF
+endif
```
Saviq: Couldn't we just add one of those to `COMMON_CONFIGURE_OPTIONS` above and only append the opposite if the conditions are met? It should be fine to override a definition later on the arguments list?

debian/rules
```diff
@@ -0,0 +1,38 @@
...
+
+
+$(info Configure options: $(COMMON_CONFIGURE_OPTIONS))
```
Saviq: That left on purpose?

tools/ppa-upload.sh
```diff
@@ -0,0 +1,148 @@
+#!/bin/bash
```
Saviq: Worried this will get out of sync with the mir version... Wonder if we need a... yeah, I know... git submodule with common CI bits?

.travis.yml
RAOF: Because otherwise the snap version of lxd detects that the non-snap version of lxd is installed and bails.

And I want to use the snap for ease of local testing, so we can always have the same lxd.

debian/rules
RAOF: Nope! Removed

debian/rules
RAOF: Yup, that (appears to) works.

debian/rules
RAOF: Nope! Boom, it's gone.

tools/ppa-upload.sh
AlanGriffiths: I agree it would be better to have any potentially shared CI integration in a separate source tree.

Does that have to be integrated via a submodule though? AIUI this would require having to sync multiple projects to the latest commit.

For CI integration it would be better to always download the current version.

spread/build/ubuntu/task.yaml
```diff
@@ -17,5 +17,5 @@ execute: |
     cd $SPREAD_PATH
     cd $(mktemp --directory)
     cmake $SPREAD_PATH
-    make -j
+    make -j4
```
Saviq: Shouldn't this be `make -j$( nproc )`?

Saviq:
bors r=AlanGriffiths,Saviq

PR #72, opened 2018-12-21
Strip generated code from git.

RAOF:
Unlike `gdbus-codegen`, where different releases have different behaviour, `wayland-scanner` is pretty sane. There's no reason not to generate the source as a part of the build process, rather than storing generated
code in git.


AlanGriffiths:
I don't have strong feelings on this, but I'm mildly against.

If something were to change in `wayland-scanner` then this will shift the point of failure from when someone intentionally regenerates the code to any time the code is regenerated as a side effect of the build process. That is a small deal for the mir team, but could be a significant inconvenience for packagers on other distros.

wmww:
I would say *not* generating automatically could cause more problems. Something that was pointed out to me is that its possible that a new libwayland could require code generated by a new wayland-scanner. I'd say that's more likely than a new newer wayland-scanner breaking interface compatibility.

PR #76, opened 2019-02-07
Run wlcs against the Mir compositor integration on build in CI

RAOF:


RAOF:
This will fail until https://github.com/MirServer/mir/pull/723 lands

RAOF:
Should work now
bors try

PR #77, opened 2019-02-08
Support dynamically marking tests as expected failures.

RAOF:
These are handled by using GTest's annotation support. A test which is
expected to fail marks itself with the "wlcs-skip-test" attribute, with
an optional skip reason, and the test listener picks up that attribute
and does not record the test it came from as a failure.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,6 +132,7 @@ set(
   src/shared_library.cpp
   src/xfail_supporting_test_listener.h
   src/xfail_supporting_test_listener.cpp
+  src/termcolor.hpp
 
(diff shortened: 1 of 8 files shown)


AlanGriffiths:
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```
bors r+

wmww:
bors r=AlanGriffiths

RAOF:
bors r=AlanGriffiths

PR #78, opened 2019-02-12
Skip tests of known-unsupported protocols.

RAOF:
This relies on tests using the `Client::acquire_interface()` method, but tests that do
will automatically be skipped if the compositor integration module indicates that
the extension, or version of the extension, is not available.


wmww:
Looks like it just has a typo

RAOF:
It does indeed just have a typo
bors r=wmww

PR #79, opened 2019-02-12
Fix program return code

RAOF:
Because `RUN_ALL_TESTS()` doesn't understand skipped tests, `wlcs` would return `EXIT_FAILURE` if a test was skipped. This is not what we want, particularly since we have skipped tests as a part of the `SelfTest` suite.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -85,8 +85,13 @@ int main(int argc, char** argv)
             listeners.Release(listeners.default_result_printer())}};
     listeners.Append(wrapping_listener);
 
-    auto result = RUN_ALL_TESTS();
-    if (result || wrapping_listener->failed())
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -85,8 +85,13 @@ int main(int argc, char** argv)
             listeners.Release(listeners.default_result_printer())}};
     listeners.Append(wrapping_listener);
 
-    auto result = RUN_ALL_TESTS();
-    if (result || wrapping_listener->failed())
+    /* (void)! is apparently the magical incantation required to get GCC to
+     * *actually* silently ignore the return value of a function declared with
+     * __attribute__(("warn_unused_result"))
+     * cf: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66425
```
wmww: C++ ❤️

Issue #84, opened 2019-03-01
Should test surface enter and leave events


wmww:
Related to https://github.com/MirServer/mir/issues/342

Saviq:
Fixed by #148

PR #85, opened 2019-03-06
Fails in Mir: test wl_surface.enter/leave events

wmww:
The surface enter half of #84

### DIFF
--- a/tests/test_surface_events.cpp
+++ b/tests/test_surface_events.cpp
@@ -588,4 +588,68 @@ TEST_F(ClientSurfaceEventsTest, buffer_release)
     EXPECT_TRUE(buffer_released[2]);
 }
 
+TEST_F(ClientSurfaceEventsTest, surface_gets_enter_event)
+{
+    using namespace testing;
+
+    wlcs::Client client{the_server()};
+    auto surface = wlcs::Surface{client};
+
+    wl_surface_listener listener = {
+        // enter
+        [](void* data, struct wl_surface*, struct wl_output*)
+            {
+                *(int*)data += 1;
+            },
+        // leave
+        [](void*, struct wl_surface*, struct wl_output*)
+            {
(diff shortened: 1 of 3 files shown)


wmww:
Now fully fixes #84

tests/test_surface_events.cpp
```diff
@@ -588,4 +588,70 @@ TEST_F(ClientSurfaceEventsTest, buffer_release)
     EXPECT_TRUE(buffer_released[2]);
 }
 
+TEST_F(ClientSurfaceEventsTest, surface_gets_enter_event)
+{
+    using namespace testing;
+
+    Test::RecordProperty("wlcs-skip-test", "Mir does not send enter events yet");
```
AlanGriffiths: We should not be implementing Mir specific limitations in these tests.

wmww: fixed

tests/test_surface_events.cpp
```diff
@@ -588,4 +588,64 @@ TEST_F(ClientSurfaceEventsTest, buffer_release)
...
+
+TEST_F(ClientSurfaceEventsTest, surface_gets_leave_event)
+{
+    using namespace testing;
+
+    wlcs::Client client{the_server()};
+    auto surface = wlcs::Surface{client};
+
+    wl_surface_listener listener = {
+        nullptr, // enter
```
RAOF: I *think* this will crash on Weston (or anything which actually supports surface enter events) because libwayland will try and call through this null pointer?

wmww: Fixed

wmww:
As we later agreed in another PR, hanging is a problem in WLCS and unrelated to any single test. Now this is just waiting on a way to disable it in Mir

wmww:
It looks like I accidentally rewrote this PR in #148. That version is better, so closing here.

PR #86, opened 2019-03-06
Fails in Mir: test for frame timestamps

### LINKED ISSUE #83: Should test surface frame timestamps
Should expose this Mir issue: https://github.com/MirServer/mir/issues/341

wmww:
Fix #83

### DIFF
--- a/tests/test_surface_events.cpp
+++ b/tests/test_surface_events.cpp
@@ -495,6 +495,53 @@ TEST_F(ClientSurfaceEventsTest, surface_moves_while_under_pointer)
         });
 }
 
+TEST_F(ClientSurfaceEventsTest, frame_timestamp_increases)
+{
+    using namespace testing;
+
+    wlcs::Client client{the_server()};
+
+    auto surface = client.create_visible_surface(100, 100);
+
+    std::array<wlcs::ShmBuffer, 3> buffers = {{
+        wlcs::ShmBuffer{client, 100, 100},
+        wlcs::ShmBuffer{client, 100, 100},
+        wlcs::ShmBuffer{client, 100, 100}
+    }};
+
+    /*
+     * The first buffer must never be released, since it is replaced before
+     * it is committed, therefore it never becomes busy.
+     */
+    wl_surface_attach(surface, buffers[0], 0, 0);
+    wl_surface_attach(surface, buffers[1], 0, 0);
+
+    int prev_frame_time = 0;
+    int frame_callback_count = 0;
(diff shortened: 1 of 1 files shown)


tests/test_surface_events.cpp
```diff
@@ -495,6 +495,52 @@ TEST_F(ClientSurfaceEventsTest, surface_moves_while_under_pointer)
         });
 }
 
+TEST_F(ClientSurfaceEventsTest, frame_timestamp_increases)
+{
+    using namespace testing;
+
+    Test::RecordProperty("wlcs-skip-test", "Mir does not send enter events yet");
```
AlanGriffiths: We should not be implementing Mir specific limitations in these tests.

wmww: fixed

tests/test_surface_events.cpp
```diff
@@ -495,6 +495,49 @@ TEST_F(ClientSurfaceEventsTest, surface_moves_while_under_pointer)
...
+    surface.add_frame_callback(
+        [&](int time)
+        {
+            EXPECT_THAT(time, Gt(prev_frame_time));
+            prev_frame_time = time;
+            frame_callback_count++;
+        });
+    wl_surface_commit(surface);
+
+    usleep(10000);
```
AlanGriffiths: Adding sleeps in the middle of tests smells badly of race conditions.

Why? And isn't there a better way?

wmww:
> I know we have other tests that hang on failure, but let's not follow a bad precedent.

WLCS hanging isn't usually the fault of bad individual tests. It's because `dispatch_until()` doesn't have a timeout, and once that gets fixed everything will stop hanging.

tests/test_surface_events.cpp
wmww: The reason is that we are testing that a millisecond resolution timestamp actually increases as time passes. Therefor we have to wait multiple milliseconds for the test to make sense.

wmww: The reason is that we are testing that a millisecond resolution timestamp actually increases as time passes. Therefor we have to wait multiple milliseconds for the test to make sense. I will add a comment explaining how it works in more detail.

wmww:
I still don't know how to disable this test in Mir

tests/test_surface_events.cpp
AlanGriffiths: It's still a shame we can't simply provide a test timesource run independently of real time.

PR #87, opened 2019-03-06
Fails in Mir: tests for subsurface reordering

### LINKED ISSUE #82: Should test subsurface reordering
These are currently not implemented in Mir, but non the less, there should be tests for them

wmww:
Fix #82

### DIFF
--- a/tests/subsurfaces.cpp
+++ b/tests/subsurfaces.cpp
@@ -474,6 +474,9 @@ TEST_P(SubsurfaceTest, place_below_simple)
 {
     auto subsurface_moving_down{wlcs::Subsurface::create_visible(main_surface, 0, 0, subsurface_width, subsurface_height)};
     wl_subsurface_place_below(subsurface_moving_down, subsurface);
+    wl_surface_commit(subsurface_moving_down);
+    wl_surface_commit(subsurface);
+    wl_surface_commit(main_surface);
 
     input_device->to_screen_position(5 + surface_x, 5 + surface_y);
     client.roundtrip();
@@ -489,6 +492,9 @@ TEST_P(SubsurfaceTest, place_above_simple)
 {
     auto subsurface_being_covered{wlcs::Subsurface::create_visible(main_surface, 0, 0, subsurface_width, subsurface_height)};
     wl_subsurface_place_above(subsurface, subsurface_being_covered);
+    wl_surface_commit(subsurface);
+    wl_surface_commit(subsurface_being_covered);
+    wl_surface_commit(main_surface);
 
     input_device->to_screen_position(5 + surface_x, 5 + surface_y);
     client.roundtrip();
--- a/tests/subsurfaces.cpp
+++ b/tests/subsurfaces.cpp
@@ -470,6 +470,36 @@ TEST_P(SubsurfaceTest, one_subsurface_to_another_fallthrough)
                     wl_fixed_from_int(pointer_y_2 - subsurface_top_y))));
 }
 
+TEST_P(SubsurfaceTest, place_below_simple)
+{
+    auto subsurface_moving_down{wlcs::Subsurface::create_visible(main_surface, 0, 0, subsurface_width, subsurface_height)};
+    wl_subsurface_place_below(subsurface_moving_down, subsurface);
+
+    input_device->to_screen_position(5 + surface_x, 5 + surface_y);
+    client.roundtrip();
+
+    ASSERT_THAT(input_device->focused_window(), Ne((wl_surface*)subsurface_moving_down))
+        << "subsurface.place_below() did not have an effect";
+
+    ASSERT_THAT(input_device->focused_window(), Ne((wl_surface*)subsurface))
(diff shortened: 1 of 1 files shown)


tests/subsurfaces.cpp
```diff
@@ -470,6 +470,40 @@ TEST_P(SubsurfaceTest, one_subsurface_to_another_fallthrough)
                     wl_fixed_from_int(pointer_y_2 - subsurface_top_y))));
 }
 
+TEST_P(SubsurfaceTest, place_below_simple)
+{
+    Test::RecordProperty("wlcs-skip-test", "Mir does not yet support subsurface reordering");
```
AlanGriffiths: We should not be implementing Mir specific limitations in these tests.

wmww: Fixed

RAOF: You've removed the `FAIL()`, but you should also remove the mark-as-skipable.

wmww:
I still don't know how I'm supposed to disable this test in Mir

RAOF:
Oh, right! I didn't quite realise what the problem to be solved was.

So, there are two obvious options here:
 * Rather than running `wlcs` against the pre-built `libmiral-wlcs-integration.so` trigger (a subset of?) Mir CI against the new WLCS. This means that Mir's test-exclusion list is authoritative, and we also test that Mir builds correctly against the new WLCS
 * Add some infrastructure to WLCS to allow integration plugins to set excluded tests.

I think the *right* solution is (1), as there's no particularly good reason for other projects to use mechanism (2); I would expect non-Mir projects to gate their CI against a wlcs *release* rather than master, so there's only this lockstep-dependency for Mir.

(1) requires that I finish a tool to trigger and wait for a Travis build from Travis, though, so (2) might be a simple choice?

AlanGriffiths:
>     * Rather than running `wlcs` against the pre-built `libmiral-wlcs-integration.so` trigger (a subset of?) Mir CI against the new WLCS. This means that Mir's test-exclusion list is authoritative, and we also test that Mir builds correctly against the new WLCS

I don't think that quite addresses the need either: we'd need to add a test exclusion to Mir every time we want to land tests to wlcs that don't (yet) pass in Mir. Having to make changes to a second project to land in this one is a Bad Idea.

My suggested option is:

1. Do not gate wlcs PRs on Mir _passing_ the tests. Provided the tests are valid, that isn't a wlcs concern (but I think the tests ought to run).

2. Mir maintains its own test _inclusion_ list that gates PRs, this can be updated as new tests become available in wlcs. New test suites would be ignored pending that update (so it would be good to have a mechanism to override this locally).

We could also have a (non-gating) Mir test run that reports which wlcs tests are failing. It would be a bonus if we could automate some "passed last time" tracking. @Saviq any thoughts?

AlanGriffiths:
After talking to RAOF we concluded that we should be able to land tests that Mir fails. But that we shouldn't be pushing wlcs development builds to the Mir PPAs. We likely need PPAs for wlcs that are independent of the Mir workflow.

If wlcs is built into its own PPAs we can manually copy the builds to the Mir PPAs when we're ready to update.

It would be nice to identify tests that Mir previously passed that fail (but that could be "two wrongs", so it shouldn't automatically gate wlcs landings).

tests/subsurfaces.cpp
```diff
@@ -470,6 +470,36 @@ TEST_P(SubsurfaceTest, one_subsurface_to_another_fallthrough)
                     wl_fixed_from_int(pointer_y_2 - subsurface_top_y))));
 }
 
+TEST_P(SubsurfaceTest, place_below_simple)
+{
+    auto subsurface_moving_down{wlcs::Subsurface::create_visible(main_surface, 0, 0, subsurface_width, subsurface_height)};
+    wl_subsurface_place_below(subsurface_moving_down, subsurface);
```
RAOF, line 476: This is actually double-buffered state, so needs a `wl_surface.commit()` on the parent before it'll be applied.

tests/subsurfaces.cpp
```diff
@@ -470,6 +470,36 @@ TEST_P(SubsurfaceTest, one_subsurface_to_another_fallthrough)
...
+        << "subsurface.place_below() did not have an effect";
+
+    ASSERT_THAT(input_device->focused_window(), Ne((wl_surface*)subsurface))
+        << "wrong surface/subsurface on top";
+}
+
+TEST_P(SubsurfaceTest, place_above_simple)
+{
+    auto subsurface_being_covered{wlcs::Subsurface::create_visible(main_surface, 0, 0, subsurface_width, subsurface_height)};
+    wl_subsurface_place_above(subsurface, subsurface_being_covered);
```
RAOF, line 491: Likewise, double-buffered state.

tests/subsurfaces.cpp
wmww: Fixed

RAOF:
bors cancel

PR #88, opened 2019-03-25
Support eventloop-based compositors

RAOF:
Many Wayland compositors are single-threaded (as libwayland is pretty difficult to use threadsafely), relying on an eventloop to do all their dispatching.

For these compositors, handling calls from arbitrary threads is difficult and annoying.

So, for these compositors, add `WlcsDisplayServer::start_on_this_thread()`. If this hook is implemented, (and `WlcsDisplayServer::start` is *not*), then WLCS will spawn a thread itself and call `start_on_this_thread`, which is expected to block until display server shutdown has completed.

When using this codepath, WLCS will proxy all calls to the compositor-under-test's mainloop, via the `wl_event_loop` passed in to `start_on_this_thread()`.


include/wlcs/display_server.h
```diff
@@ -25,8 +25,9 @@
 extern "C" {
 #endif
 
-typedef struct wl_surface wl_surface;
-typedef struct wl_display wl_display;
+struct wl_surface;
+struct wl_display;
+struct wl_event_loop;
```
AlanGriffiths: Why this change of style? Unless there's a good reason I'd recommend not doing this: The original `typedef struct wl_surface wl_surface` (as used elsewhere in the file) introduces `wl_surface`` to both C namespaces, and hides that incompatibility with C++.

src/in_process_server.cpp
```diff
@@ -35,6 +35,407 @@
...
+struct callable_traits<Returns(Callable::*)(Args...) const>
+{
+    using return_type = Returns;
+    using args = typename std::tuple<Args...>;
+};
+
+/*
+ * call_helper and call are downmarket versions of C++17's std::invoke()
+ *
+ * We don't yet require C++17, so downmarket it is!
```
AlanGriffiths: Maybe we should? I don't think we would have a problem on any supported OS.

AlanGriffiths: Although... on 16.04LTS, g++5.4 lacks `invoke()` from `<functional>` despite accepting `-std=c++17`.

include/wlcs/display_server.h
RAOF: So, there's an idiom in Wayland to have code like
```
    struct wl_surface* wl_surface;
```
Which IIRC breaks if we have the typedef visible.

src/thread_proxy.h
```diff
@@ -0,0 +1,468 @@
+/*
+ * Copyright © 2017 Canonical Ltd.
```
AlanGriffiths: 2017?!

include/wlcs/display_server.h
AlanGriffiths: OK, that's a fair justification. (Although I'd argue it is an "unrelated change" and ought to be done consistently.)

RAOF:
That should be everything resolved?

AlanGriffiths:
> # Timed out

bors r+

Saviq:
Travis seems to have gotten stuck here: https://travis-ci.org/MirServer/wlcs/builds/512536650

Canceled and trying again.

bors r=AlanGriffiths

Saviq:
And stuck again… Travis unhappy today.

bors r=AlanGriffiths

Saviq:
OK Increased the timeout now…

bors r=AlanGriffiths

RAOF:
Once more, with feeling:
bors r=AlanGriffiths

RAOF:
bors r=AlanGriffiths

RAOF:
bors ping

RAOF:
bors r=AlanGriffiths

RAOF:
That should fix the conflicts. Now Travis has decided to be slow, but I'll give this another try:

bors r=AlanGriffiths

PR #90, opened 2019-03-29
Fix signature of WlcsDisplayServer::get_descriptor

RAOF:
It's entirely possible that the set of extensions a compositor supports
might be different with different invokations (eg: Mir's default_extensions).
We shouldn't require the set of extensions to be fixed at compositor-integration
time, but allow them to be WlcsDisplayServer-specific.


RAOF:
bors try

RAOF:
bors r=wmww

RAOF:
Is this a branch-protection issue?
bors r=wmww

PR #91, opened 2019-04-26
Release 1.0

RAOF:


RAOF:
I don't think we particularly need a release testing plan, other than CI.

We *do* need to cherry-pick merge #92 before CI will pass, though.

PR #93, opened 2019-05-01
wl_output test

wmww:


include/in_process_server.h
```diff
@@ -177,6 +177,16 @@ class ShmBuffer
     std::unique_ptr<Impl> impl;
 };
 
+struct OutputState
+{
+    wl_output *output;
+    int width = -1, height = -1;
+    int scale = -1;
```
RAOF: This looks like a job for `std::optional<int> width, height, scale;`?

src/in_process_server.cpp
```diff
@@ -1166,6 +1251,17 @@ class wlcs::Client::Impl
             me->data_device_manager = static_cast<struct wl_data_device_manager*>(
                 wl_registry_bind(registry, id, &wl_data_device_manager_interface, version));
         }
+        else if ("wl_output"s == interface)
+        {
+            auto wl_output = static_cast<struct wl_output*>(
+                wl_registry_bind(registry, id, &wl_output_interface, version));
+            auto output = std::make_unique<Output>(wl_output);
+            wl_output_add_listener(wl_output, &Output::listener, output.get());
+            me->outputs.push_back(move(output));
```
RAOF: We will, at some point, need to handle the `wl_output`s going away in `global_remove`, but that can be for later.

tests/wl_output.cpp
```diff
@@ -0,0 +1,43 @@
...
+using namespace testing;
+using namespace wlcs;
+
+using WlOutputTest = wlcs::InProcessServer;
+
+TEST_F(WlOutputTest, wl_output_properties_set)
+{
+    wlcs::Client client{the_server()};
+
+    ASSERT_THAT(client.output_count(), Eq(1u));
```
RAOF: I think this is currently to-specific a test; as we don't have an interface on `WlcsDisplayServer` to request the addition/removal of mock displays, I don't think we should currently be (implicitly!) requiring that a fixture provides exactly one display.

Make this ≥ 1; it's not sensible to have *no* `wl_output`…

tests/wl_output.cpp
```diff
@@ -0,0 +1,43 @@
...
+    auto output = client.output_state(0);
+
+    EXPECT_THAT(output.geometry_set, Eq(true));
+    EXPECT_THAT(output.mode_set, Eq(true));
+    EXPECT_THAT(output.scale_set, Eq(true));
+
+    EXPECT_THAT(output.scale, Eq(1));
```
RAOF: …and then, likewise, is it sensible to say that the scale of the `wl_output` *must* be 1, without having specified it?

I'd probably just remove this bit (unless and until you want to add a mock output configuration interface to `WlcsDisplayServer`; feel free to do that!)

include/in_process_server.h
```diff
@@ -201,6 +211,9 @@ class Client
     Surface create_xdg_shell_stable_surface(int width, int height);
     Surface create_visible_surface(int width, int height);
 
+    size_t output_count() const;
+    OutputState output_state(size_t index) const;
```
RAOF: I wish it were less of an absolute pain in the arse to provide iterators in C++. This is an crying out to be an iterator.

AlanGriffiths:
@RAOF have your issues been addressed?

RAOF:
Yup, I'm good with this now.

bors r+

PR #106, opened 2019-06-28
Basic tests for primary selection device

AlanGriffiths:
These SKIP with miral_wlcs_integration.so
They PASS with the egmde "primary-selection" branch (that currently needs the Mir "primary-selection" branch to build).

Almost certainly we will need more tests, but this is a start...


AlanGriffiths:
I'd check whether the "zwp_primary_selection_device_manager_v1" interface was "supported", _if I could figure out #107_!

include/in_process_server.h
```diff
@@ -206,6 +207,7 @@ class Client
     wl_shm* shm() const;
     wl_data_device_manager* data_device_manager() const;
     wl_seat* seat() const;
+    struct zwp_primary_selection_device_manager_v1* primary_selection_device_manager() const;
```
wmww: Non-blocking but you don't need the `struct` keyword, and I'm trying to move away from using it in this case.

AlanGriffiths:
bors r=wmww

PR #111, opened 2019-07-03
dispatch_until(): Timeout rather than waiting indefinitely.

RAOF:
Do the (surprisingly involved) dance required to `poll` the wayland fd so that we can wait for a specified time, rather than forever.

Fixes: #89

### DIFF
--- a/tests/self_test.cpp
+++ b/tests/self_test.cpp
@@ -150,3 +150,23 @@ TEST_F(SelfTest, acquiring_unsupported_extension_version_is_xfail)
 
     FAIL() << "We should have (x)failed at acquiring the interface";
 }
+
+TEST_F(SelfTest, dispatch_until_times_out_on_failure)
+{
+    Client client{the_server()};
+
+    // Ensure that there's some events happening on the Wayland socket
+    auto dummy = client.create_visible_surface(300, 300);
+    dummy.attach_buffer(300, 300);
+    wl_surface_commit(dummy);
+
+    try
+    {
+        client.dispatch_until([]() { return false; });
+    }
+    catch (wlcs::Timeout const&)
+    {
+        return;
+    }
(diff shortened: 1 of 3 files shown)


tests/self_test.cpp
```diff
@@ -150,3 +150,44 @@ TEST_F(SelfTest, acquiring_unsupported_extension_version_is_xfail)
...
+        return;
+    }
+    FAIL() << "Dispatch did not raise a wlcs::Timeout exception";
+}
+
+TEST_F(SelfTest, dispatch_until_times_out_at_the_right_time)
+{
+    Client client{the_server()};
+
+    auto const timeout = std::chrono::seconds{10};
```
AlanGriffiths: I know builders can be slow, but this seems excessive.

tests/self_test.cpp
```diff
@@ -150,3 +150,44 @@ TEST_F(SelfTest, acquiring_unsupported_extension_version_is_xfail)
...
+    try
+    {
+        client.dispatch_until([]() { return false; }, timeout);
+    }
+    catch (wlcs::Timeout const&)
+    {
+        auto const skew = std::chrono::steady_clock::now() - expected_end;
+        EXPECT_THAT(
+            abs(std::chrono::duration_cast<std::chrono::seconds>(skew).count()),
+            Le(5) );
```
AlanGriffiths: This seems wrong. I'd expect something like:
```c++
        EXPECT_THAT(skew, Ge(0s));
        EXPECT_THAT(skew, Lt(1s));
```

AlanGriffiths: Or even:
```c++
        EXPECT_THAT(std::chrono::steady_clock::now(), Ge(expected_end));
        EXPECT_THAT(std::chrono::steady_clock::now(), Lt(expected_end+5s));
```

PR #116, opened 2019-07-16
Test layer shell protocol

wmww:
Adds tests for the Layer Shell protocol. Mir passes with https://github.com/MirServer/mir/pull/926.

### DIFF
--- a/include/in_process_server.h
+++ b/include/in_process_server.h
@@ -137,6 +137,7 @@ public:
 
     void attach_buffer(int width, int height);
     void add_frame_callback(std::function<void(int)> const& on_frame);
+    void attach_visible_buffer(int width, int height);
 
     bool has_focus() const;
     std::pair<int, int> pointer_position() const;
--- a/include/in_process_server.h
+++ b/include/in_process_server.h
@@ -40,6 +40,8 @@ struct WlcsPointer;
 struct WlcsTouch;
 struct WlcsServerIntegration;
 
+struct zwlr_layer_shell_v1;
+
 namespace wlcs
 {
 
@@ -225,6 +227,7 @@ public:
     wl_shell* shell() const;
     zxdg_shell_v6* xdg_shell_v6() const;
     xdg_wm_base* xdg_shell_stable() const;
+    zwlr_layer_shell_v1* layer_shell_v1() const;
     wl_surface* focused_window() const;
     wl_surface* touched_window() const;
     std::pair<wl_fixed_t, wl_fixed_t> pointer_position() const;
--- a/include/layer_shell_v1.h
+++ b/include/layer_shell_v1.h
@@ -1,0 +1,68 @@
+/*
+ * Copyright © 2019 Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: William Wold <william.wold@canonical.com>
+ */
+
+#ifndef WLCS_LAYER_SHELL_V1_H
+#define WLCS_LAYER_SHELL_V1_H
+
+#include "in_process_server.h"
+
+// Because _someone_ *cough*ddevault*cough* thought it would be a great idea to name an argument "namespace"
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wkeyword-macro"
+#endif
+#define namespace _namespace
+#include "generated/wlr-layer-shell-unstable-v1-client.h"
+#undef namespace
+#ifdef __clang__
+#pragma clang diagnostic pop
+#endif
+
+namespace wlcs
+{
+
+class LayerSurfaceV1
+{
+public:
+    LayerSurfaceV1(
+        wlcs::Client& client,
+        wlcs::Surface& surface,
+        zwlr_layer_shell_v1_layer layer = ZWLR_LAYER_SHELL_V1_LAYER_TOP,
+        wl_output* output = NULL,
+        const char *_namespace = "wlcs");
+    LayerSurfaceV1(LayerSurfaceV1 const&) = delete;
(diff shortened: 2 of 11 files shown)


tests/wlr_layer_shell_v1.cpp
```diff
@@ -234,6 +225,79 @@ class LayerSurfaceAnchorTest:
...
+        os << "Layer{";
+        if (i)
+        {
+            switch (i.value())
+            {
+            case ZWLR_LAYER_SHELL_V1_LAYER_BACKGROUND:
+                os << "background";
+                break;
+            case ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM:
+                os << "background";
```
AlanGriffiths: It looks wrong to write the same text for all layers.

wmww: Good catch

include/layer_shell_v1.h
```diff
@@ -0,0 +1,68 @@
...
+    operator zwlr_layer_surface_v1*() const { return layer_surface; }
+
+    void dispatch_until_configure();
+    auto last_width() const -> int { return _last_width; }
+    auto last_height() const -> int { return _last_height; }
+
+private:
+    zwlr_layer_surface_v1* layer_surface;
+    wlcs::Client& client;
+    int _last_width = -1;
```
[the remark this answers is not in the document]
wmww: Ack

src/in_process_server.cpp
```diff
@@ -806,6 +796,14 @@ class wlcs::Client::Impl
         return xdg_shell_stable;
     }
 
+    zwlr_layer_shell_v1* the_layer_shell_v1() const
+    {
+        if (layer_shell_v1)
+            return layer_shell_v1;
+        else
+            BOOST_THROW_EXCEPTION((ExtensionExpectedlyNotSupported{"zwlr_layer_shell_v1", 0}));
```
AlanGriffiths: This is wrong.

`ExtensionExpectedlyNotSupported` is used elsewhere to indicate that the extension is not in the `supported_extensions`provided by the test fixture.

Here we've discovered that the server doesn't publish `layer_shell_v1` as a global, but we are not checking `supported_extensions`. IMO the test for `supported_extensions` should be earlier: if we get to here then we should be failing.

See, for example, the use of `CheckInterfaceExpected` in recent test fixtures.

include/layer_shell_v1.h
```diff
@@ -56,6 +56,7 @@ class LayerSurfaceV1
     auto last_height() const -> int { return last_height_; }
 
 private:
+    CheckInterfaceExpected expect_layer_shell_supported;
```
AlanGriffiths: I'm not convinced this test belongs in `wlcs::LayerSurfaceV1` and not explicitly in each of the fixtures or tests that use it.

In any case, it doesn't need to be a member variable - it can equally be local to the constructor.

include/in_process_server.h
```diff
@@ -254,6 +254,8 @@ class Client
         std::chrono::seconds timeout = std::chrono::seconds{10});
 
     void roundtrip();
+
+    auto server() const -> wlcs::Server&;
```
AlanGriffiths: Given my comment below, I'm unconvinced that there's a good motivation for adding this.

Maybe, what is needed is a "check_interface_expected()" method?

wmww: whether or not it's needed here, I think it's useful to have. I'll pull it out of this PR though.

wmww:
bors cancel

include/in_process_server.h
AlanGriffiths: I think it likely to lead to a [train wreck](https://devcards.io/train-wreck).

wmww: Ehh... that's fair. Perhaps Client should hold a server reference, but not expose it directly. Either way, I'll get rid of this for now.

src/layer_shell_v1.cpp
```diff
@@ -0,0 +1,69 @@
...
+    : client{client}
+{
+    layer_surface = zwlr_layer_shell_v1_get_layer_surface(
+        client.layer_shell_v1(),
+        surface,
+        output,
+        layer,
+        _namespace);
+    static struct zwlr_layer_surface_v1_listener const listener {
+        +[](void* data,
```
RAOF: What's the "+" prefix doing here? Is this to force decay to a function pointer?

tests/wlr_layer_shell_v1.cpp
```diff
@@ -0,0 +1,564 @@
...
+
+    void expect_surface_is_at_position(std::pair<int, int> pos)
+    {
+        auto pointer = the_server().create_pointer();
+        pointer.move_to(pos.first + 2, pos.second + 3);
+        client.roundtrip();
+
+        ASSERT_THAT(client.focused_window(), Eq((wl_surface*)surface));
+        ASSERT_THAT(wl_fixed_to_int(client.pointer_position().first), Eq(2));
+        ASSERT_THAT(wl_fixed_to_int(client.pointer_position().second), Eq(3));
```
RAOF: I generally prefer `EXPECT_THAT` over `ASSERT_THAT` for expectations that are not preconditions of later expectations.

In this case, `x != 2` does not mean we can't check the value of `y`, so these should be `EXPECT_THAT`, right?

src/layer_shell_v1.cpp
wmww: Correct

tests/wlr_layer_shell_v1.cpp
wmww: Agreed

wmww:
bors try

wmww:
bors try

src/layer_shell_v1.cpp
AlanGriffiths: The decay happens anyway, I don't think the "+" makes anything clearer to the reader.

include/layer_shell_v1.h
wmww: This has been resolved

src/layer_shell_v1.cpp
wmww: Strange, I thought those pluses were required. Maybe they were in C++11 or something. Anyway, removed.

include/layer_shell_v1.h
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * Authored by: William Wold <william.wold@canonical.com>
+ */
+
+#ifndef WLCS_LAYER_SHELL_V1_H
+#define WLCS_LAYER_SHELL_V1_H
+
+#include "in_process_server.h"
+
+// Because _someone_ *cough*ddevault*cough* thought it would be a great idea to name an argument "namespace"
```
AlanGriffiths: Just maybe it is the generator at fault, why should a protocol definition care what the keywords are in all possible implementation languages?

AlanGriffiths:
Seems spread trouble has now passed.

bors r+

PR #124, opened 2019-07-25
Add test for buffer attached before first commit

wmww:
Currently fails on Mir due to https://github.com/MirServer/mir/issues/936


AlanGriffiths:
> I think this one is _not_ a WLCS test. AFAIK there's nothing in the protocol which states that newly realised windows must gain input focus, so this is testing a Mir-specific behaviour rather than protocol conformance.

Agreed. It isn't even a Mir-specific behaviour, just a default inherited from the Unity8 design.

Even testing that the buffer is rendered goes beyond what the protocol requires.

PR #125, opened 2019-08-02
Fix Client::dispatch_until() timing out early

RAOF:
Because `duration_cast<milliseconds>(maximum_wait)` performs integer division we lose any fractional milliseconds in `maximum_wait`, meaning that `poll()` could return up to 1ms before we were intending to timeout.

### DIFF
--- a/include/gtest_helpers.h
+++ b/include/gtest_helpers.h
@@ -1,0 +1,46 @@
+/*
+ * Copyright © 2019 Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 3,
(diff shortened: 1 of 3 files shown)


RAOF:
Bors try

RAOF:
Bors try

Let's give it a second chance at failing

RAOF:
That was a Fedora setup issue...

bors try

AlanGriffiths:
That was a Fedora setup issue...

bors try

RAOF:
Ok. I'm *reasonably* certain that this now *actually* fixes the problem.

AlanGriffiths:
LGTM

bors r+

PR #126, opened 2019-08-02
Remove extra implementations of StartedInProcessServer

wmww:
These classes are both redundant and incorrect. The better implementation in `in_process_server.h` which these hide calls `InProcessServer::TearDown()` in the destructor. Not doing this was, for some reason, causing some Wayland connections to not close (as observed by wayland-debug)


wmww:
bors r=AlanGriffiths

wmww:
bors r=AlanGriffiths

PR #128, opened 2019-08-07
Revert "Allow Fedora 30 to fail (while there are setup issues)"

AlanGriffiths:
This reverts commit 9f7230e3


AlanGriffiths:
bors r=RAOF

PR #132, opened 2019-08-15
Fix FTBFS with packaging options

AlanGriffiths:
Can be tested using debuild etc.


RAOF:
Internal Travis network problems.

bors r=wmww

PR #133, opened 2019-08-16
Improve touch and pointer tests

wmww:
* Ensure touch and pointer events are followed by a `.frame`
* Better (but not complete) support for multiple touches
* Check for various invalid events
  * The `Got wl_pointer.leave with the wrong surface` check exposed https://github.com/MirServer/mir/issues/968, so that needs to be fixed
* Test touches on subsurfaces with an offset
  * Confirmed a bug I'd observed in manual testing: https://github.com/MirServer/mir/issues/967
* Test touches on XDG toplevels with window geometry set
  * Also causes https://github.com/MirServer/mir/issues/967

### DIFF
--- a/include/in_process_server.h
+++ b/include/in_process_server.h
@@ -236,6 +236,7 @@ public:
     std::pair<wl_fixed_t, wl_fixed_t> pointer_position() const;
     std::pair<wl_fixed_t, wl_fixed_t> touch_position() const;
     bool pointer_events_clean() const; ///< If we have gotten a frame event since the last pointer event
+    bool touch_events_clean() const; ///< If we have gotten a frame event since the last touch event
 
     using PointerEnterNotifier =
         std::function<bool(wl_surface*, wl_fixed_t x, wl_fixed_t y)>;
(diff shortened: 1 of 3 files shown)


wmww:
Turns out if a Wayland object is destroyed while a message using that object is inflight, libwayland kindly nulls out that message argument. This was the cause of MirServer/mir#968, and the solution is simply to accept nulls in `wl_pointer.leave`. All the Mir issues should now be resolved.

wmww:
Strange to be only missing on 16.04, but that's what it appears to be

AlanGriffiths:
> Strange to be only missing on 16.04, but that's what it appears to be

Not so strange: std headers are allowed, but not required, to include each other. (An "interesting" choice by the library committee, but that's what it is.)

include/in_process_server.h
```diff
@@ -235,6 +235,8 @@ class Client
     wl_surface* touched_window() const;
     std::pair<wl_fixed_t, wl_fixed_t> pointer_position() const;
     std::pair<wl_fixed_t, wl_fixed_t> touch_position() const;
+    bool pointer_events_clean() const; ///< If we have gotten a frame event since the last pointer event
```
AlanGriffiths: The name `pointer_events_clean()` doesn't say much to me (compared to, say, `!pointer_events_pending()`).

But is this even used?

PR #135, opened 2019-09-07
Remove unused code that uses dlvsym()

z3ntu:
Mostly copied from Mir.

### DIFF
--- a/src/shared_library.cpp
+++ b/src/shared_library.cpp
@@ -53,15 +53,3 @@ void* wlcs::SharedLibrary::load_symbol(char const* function_name) const
         BOOST_THROW_EXCEPTION(std::runtime_error(dlerror()));
     }
 }
-
-void* wlcs::SharedLibrary::load_symbol(char const* function_name, char const* version) const
-{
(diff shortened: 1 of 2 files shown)


src/shared_library.cpp
```diff
@@ -56,6 +56,14 @@ void* wlcs::SharedLibrary::load_symbol(char const* function_name) const
 
 void* wlcs::SharedLibrary::load_symbol(char const* function_name, char const* version) const
 {
+    // dlvsym is a GNU extension and does not exist on non-glibc (e.g. musl)
+#ifdef WLCS_DONT_USE_DLVSYM
+    // Load the function without checking the version
+    //log_debug("Cannot check \"%s\" symbol version is \"%s\": dlvsym() is unavailable", function_name, version);
```
z3ntu: This logging function doesn't exist in wlcs, should I just remove the logging?

z3ntu:
Updated!

RAOF:
Looks good. The commit message is now a bit misleading, though 😀.

If you don't come up with better wording I'll do it and merge this when I'm back at work tomorrow.

PR #139, opened 2019-09-12
Update debian/watch file to point to GitHub releases

RAOF:


Saviq:
In that case:

bors r+

PR #140, opened 2019-09-12
Travis: Use raw LXD, rather than farming out to spread.

RAOF:
We don't really use any of the features of spread, have to rely on a
fork of spread that supports LXD and Fedora.

Let's not rely on my home server being up for CI to work!


RAOF:
We shall continue on with spread!

PR #141, opened 2019-09-16
focused_window -> pointed_at_window

wmww:
window "focus" can mean several things, and "has the pointer over it" is rarely one of them.


wmww:
I agree that's a better name

PR #143, opened 2019-09-30
Test wl_output.release

wmww:
Tests that the `wl_output.release` request doesn't cause the compositor any trouble (regression test for https://github.com/MirServer/mir/pull/1015)

### DIFF
--- a/src/in_process_server.cpp
+++ b/src/in_process_server.cpp
@@ -1626,6 +1626,19 @@ wlcs::OutputState wlcs::Client::output_state(size_t index) const
     return impl->outputs[index]->current;
 }
 
+void wlcs::Client::release_output(size_t index)
+{
+    if (index > output_count())
+        throw std::runtime_error("Invalid output index");
+
+    auto const version = wl_output_get_version(impl->outputs[index]->current.output);
+    if (version < WL_OUTPUT_RELEASE_SINCE_VERSION)
+        throw std::runtime_error("Output version " + std::to_string(version) + " too low to support .release");
+
+    wl_output_release(impl->outputs[index]->current.output);
+    impl->outputs.erase(impl->outputs.begin() + index);
+}
+
 wl_shell* wlcs::Client::shell() const
 {
     return impl->the_shell();
--- a/src/in_process_server.cpp
+++ b/src/in_process_server.cpp
@@ -1621,7 +1621,7 @@ size_t wlcs::Client::output_count() const
 wlcs::OutputState wlcs::Client::output_state(size_t index) const
 {
     if (index > output_count())
-        throw std::runtime_error("Invalid output index");
+        throw std::out_of_range("Invalid output index");
 
     return impl->outputs[index]->current;
 }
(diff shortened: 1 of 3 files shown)


src/in_process_server.cpp
```diff
@@ -1626,6 +1626,19 @@ wlcs::OutputState wlcs::Client::output_state(size_t index) const
     return impl->outputs[index]->current;
 }
 
+void wlcs::Client::release_output(size_t index)
+{
+    if (index > output_count())
+        throw std::runtime_error("Invalid output index");
```
RAOF: Nit: `throw std::out_of_range("Invalid output index");`

src/in_process_server.cpp
```diff
@@ -1626,6 +1626,19 @@ wlcs::OutputState wlcs::Client::output_state(size_t index) const
...
+
+    auto const version = wl_output_get_version(impl->outputs[index]->current.output);
+    if (version < WL_OUTPUT_RELEASE_SINCE_VERSION)
+        throw std::runtime_error("Output version " + std::to_string(version) + " too low to support .release");
```
RAOF: Nit: This is probably a `std::logic_error`

tests/wl_output.cpp
```diff
@@ -39,3 +39,15 @@ TEST_F(WlOutputTest, wl_output_properties_set)
...
+{
+    wlcs::Client client{the_server()};
+
+    // Skip test if compositor doesn't support a new enough wl_output version
+    client.acquire_interface(wl_output_interface.name, &wl_output_interface, WL_OUTPUT_RELEASE_SINCE_VERSION);
+
+    ASSERT_THAT(client.output_count(), Ge(1u));
+    client.release_output(0);
+    client.roundtrip();
+}
```
RAOF: I think this would be clearer as:
```
// Acquire *any* wl_output; we don't care which
auto output = static_cast<wl_output*>(client.acquire_interface(wl_output_interface.name, &wl_output_interface, WL_OUTPUT_RELEASE_SINCE_VERSION));
wl_output_release(output);
client.roundtrip()
```

You could then drop the `client.release_output()` implementation entirely.

wmww: That's what I tried first. The problem is that since `wl_output.release` is marked as a destructor in the protocol, it is invalid to `wl_output_destroy()` it after a release (it segfaults). Thus the `client.release` which removes the output from the list and prevents it from being destroyed later.

RAOF: But `client.acquire_interface(...)` returns a *new* `wl_resource` that won't be freed by any other code; nothing will attempt to call `wl_output_destroy()` on it.

wmww: Good point. Done

wmww:
bors r=RAOF

RAOF:
Fedora networking!!!
Bors r+

AlanGriffiths:
> Fedora networking!!!

Again.

bors r+

PR #144, opened 2019-10-04
XDG Output test

wmww:
This will become more useful once we have WLCS automatically testing all protocol versions

### DIFF
--- a/src/xdg_output_v1.cpp
+++ b/src/xdg_output_v1.cpp
@@ -19,12 +19,50 @@
 #include "xdg_output_v1.h"
 #include <boost/throw_exception.hpp>
 
+struct wlcs::XdgOutputManagerV1::Impl
+{
+    Impl(Client& client)
+        : client{client},
+          manager{static_cast<struct zxdg_output_manager_v1*>(
+              client.acquire_interface(
+                  zxdg_output_manager_v1_interface.name,
+                  &zxdg_output_manager_v1_interface,
+                  zxdg_output_manager_v1_interface.version))}
+    {
+    }
+
+    ~Impl()
+    {
+        zxdg_output_manager_v1_destroy(manager);
+    }
+
+    Client& client;
+    zxdg_output_manager_v1* const manager;
+};
+
+wlcs::XdgOutputManagerV1::XdgOutputManagerV1(Client& client)
+    : impl{std::make_unique<Impl>(client)}
+{
+}
+
+wlcs::XdgOutputManagerV1::~XdgOutputManagerV1() = default;
+
+wlcs::XdgOutputManagerV1::operator zxdg_output_manager_v1*() const
+{
+    return impl->manager;
+}
+
+auto wlcs::XdgOutputManagerV1::client() const -> Client&
+{
+    return impl->client;
+}
+
 struct wlcs::XdgOutputV1::Impl
 {
-    Impl(Client& client, size_t output_index)
+    Impl(XdgOutputManagerV1& manager, size_t output_index)
         : output{zxdg_output_manager_v1_get_xdg_output(
-              client.xdg_output_manager_v1(),
-              client.output_state(output_index).output)},
+              manager,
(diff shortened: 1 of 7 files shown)


src/xdg_output_v1.cpp
```diff
@@ -0,0 +1,113 @@
...
+        impl->dirty = true;
+    },
+};
+
+wlcs::XdgOutputV1::XdgOutputV1(Client& client, size_t output_index)
+    : impl{std::make_shared<Impl>(zxdg_output_manager_v1_get_xdg_output(
+          client.xdg_output_manager_v1(),
+          client.output_state(output_index).output))}
+{
+    zxdg_output_v1_add_listener(impl->output, &Impl::listener, impl.get());
```
RAOF: This rather depends on WLCS implementation details, doesn't it? Adding the listener here relies on WLCS core not having already dispatched (into the void) the initial configuration events, right? (This is another thing that using `client.acquire_interface()` would avoid)

src/xdg_output_v1.cpp
```diff
@@ -0,0 +1,113 @@
...
+        auto const impl = static_cast<Impl*>(data);
+        impl->_state.logical_size = std::make_pair(width, height);
+        impl->dirty = true;
+    },
+
+    [] /* done */ (void *data, zxdg_output_v1 *)
+    {
+        auto const impl = static_cast<Impl*>(data);
+        if (impl->version < 3)
+            impl->dirty = false;
```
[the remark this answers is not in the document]
wmww: > For objects version 3 onwards, this event is deprecated. Compositors are not required to send it anymore and must send wl_output.done instead.

So no, because it's not forbidden for compositors to send it

src/xdg_output_v1.cpp
wmww: It's not adding a listener to the `zxdg_output_manager_v1` (which is managed by WLCS), it's adding a listener to the `zxdg_output_v1` (which is created a few lines above and doesn't leave the file).

wmww: I moved this into the `Impl` constructor to make it more obvious

src/xdg_output_v1.cpp
```diff
@@ -0,0 +1,113 @@
...
+    {
+        auto const impl = static_cast<Impl*>(data);
+        impl->_state.description = description;
+        impl->dirty = true;
+    },
+};
+
+wlcs::XdgOutputV1::XdgOutputV1(Client& client, size_t output_index)
+    : impl{std::make_shared<Impl>(zxdg_output_manager_v1_get_xdg_output(
+          client.xdg_output_manager_v1(),
```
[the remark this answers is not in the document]
wmww: Done. I think a lot of stuff can probably be moved from `in_process_server.cpp` to their own files. I just did it this way because I it was how everything else was working.

src/xdg_output_v1.cpp
wmww: I moved this into the Impl constructor to make it more obvious

RAOF:
Heh, maybe a spec bug is in order 😀

I interpret “must send wl_output.done instead” as forbidding the sending of xdg_output.done. That is, sending done on a v3 object is an error.

wmww:
But is that enough to overrule "compositors are not required to send it"? The protocol had an opportunity to say "compositors must not send it", but did not choose those words.

tests/xdg_output_v1.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include <gmock/gmock.h>
+
+using namespace testing;
+using namespace wlcs;
+
+class XdgOutputV1Test
+    : public wlcs::StartedInProcessServer
+{
+private:
+    wlcs::CheckInterfaceExpected expect_xdg_output_supported{the_server(), zxdg_output_manager_v1_interface};
```
RAOF: You no longer need this; the `acquire_interface` call in `XdgOutputManagerV1` does the same job.

PR #145, opened 2019-10-15
Fails in Mir: Test that subsurface moves under input device

wmww:
I wrote this to expose a problem I found in Mir. Will reference this PR in the Mir issue

### DIFF
--- a/tests/subsurfaces.cpp
+++ b/tests/subsurfaces.cpp
@@ -508,6 +508,70 @@ TEST_P(SubsurfaceTest, subsurface_of_a_subsurface_handled)
                     wl_fixed_from_int(pointer_y_2 - subsurface_top_y - subsurface_y))));
 }
 
+TEST_P(SubsurfaceTest, subsurface_moves_under_input_device_once)
+{
+    int const input_x = surface_x + 10, input_y = surface_y + 5;
+    int const subsurface_x = -23, subsurface_y = -17;
+
+    input_device->to_screen_position(input_x, input_y);
+    client.roundtrip();
+
+    ASSERT_THAT(input_device->focused_window(), Eq((wl_surface*)subsurface)) << "precondition failed";
+    ASSERT_THAT(input_device->position_on_window(),
+                Eq(std::make_pair(
+                    wl_fixed_from_int(input_x - surface_x),
+                    wl_fixed_from_int(input_y - surface_y)))) << "precondition failed";
+
+    wl_subsurface_set_position(subsurface, subsurface_x, subsurface_y);
+    wl_surface_commit(subsurface);
+    wl_surface_commit(main_surface);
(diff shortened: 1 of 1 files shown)


tests/subsurfaces.cpp
```diff
@@ -508,6 +508,70 @@ TEST_P(SubsurfaceTest, subsurface_of_a_subsurface_handled)
...
 
+TEST_P(SubsurfaceTest, subsurface_moves_under_input_device_once)
+{
+    int const input_x = surface_x + 10, input_y = surface_y + 5;
+    int const subsurface_x = -23, subsurface_y = -17;
+
+    input_device->to_screen_position(input_x, input_y);
+    client.roundtrip();
+
+    ASSERT_THAT(input_device->focused_window(), Eq((wl_surface*)subsurface)) << "precondition failed";
```
RAOF: Non-blocking nit: I don't think “precondition failed” is a useful label for `ASSERT_THAT`. It already appears in the output, and we only use `ASSERT_THAT` for precondition failures.

tests/subsurfaces.cpp
```diff
@@ -508,6 +508,70 @@ TEST_P(SubsurfaceTest, subsurface_of_a_subsurface_handled)
...
+                Ne(std::make_pair(
+                    wl_fixed_from_int(input_x - surface_x),
+                    wl_fixed_from_int(input_y - surface_y)))) << "input device did not get new location";
+    EXPECT_THAT(input_device->position_on_window(),
+                Eq(std::make_pair(
+                    wl_fixed_from_int(input_x - surface_x - subsurface_x),
+                    wl_fixed_from_int(input_y - surface_y - subsurface_y)))) << "input device in wrong location";
+}
+
+TEST_P(SubsurfaceTest, subsurface_moves_under_input_device_twice)
```
RAOF: Hm. Why do we have this test? Is there any reason to expect that, after the first one passes this one might fail?

wmww: I don't remember exactly why I added this. My vague recollection is that it tested something I noticed was broken in Mir.

tests/subsurfaces.cpp
wmww: > We only use `ASSERT_THAT` for precondition failures

While this is generally my policy, I'm not sure it is used consistently throughout Mir. It's also easy to miss that the failed thing is an assert and not expect. Especially if someone isn't used to gtest or our way of using gtest, immediate seeing "precondition failed" right in the output might save them some time.

tests/subsurfaces.cpp
wmww: It appears in current Mir master, `subsurface_moves_under_input_device_twice` fails in all cases while once only fails with touch

wmww:
bors r=RAOF

PR #146, opened 2019-10-17
XDG toplevel parent tests

wmww:
Currently fails in Mir. PR on the way.

### DIFF
--- a/tests/xdg_toplevel_stable.cpp
+++ b/tests/xdg_toplevel_stable.cpp
@@ -222,6 +222,36 @@ TEST_F(XdgToplevelStableTest, interactive_move)
     client.roundtrip();
 }
 
+TEST_F(XdgToplevelStableTest, parent_can_be_set)
+{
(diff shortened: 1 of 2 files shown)


tests/xdg_toplevel_stable.cpp
```diff
@@ -222,6 +222,36 @@ TEST_F(XdgToplevelStableTest, interactive_move)
...
+
+    ConfigurationWindow parent{client};
+    the_server().move_surface_to(parent.surface, window_pos_x, window_pos_y);
+
+    ConfigurationWindow child{client};
+    the_server().move_surface_to(child.surface, window_pos_x, window_pos_y);
+
+    xdg_toplevel_set_parent(child, parent);
+    wl_surface_commit(child.surface);
+    client.roundtrip();
```
RAOF: Hm. Future work: it would be nice if we could describe this as `EXPECT_NO_PROTOCOL_ERROR()` or something.

PR #148, opened 2020-02-28
[FAILS ON MIR] Test surfaces get enter/leave events

wmww:
I'm using these tests to implement the enter/leave events on Mir, which are needed for HiDPI support

### DIFF
--- a/src/in_process_server.cpp
+++ b/src/in_process_server.cpp
@@ -1821,30 +1821,30 @@ private:
     {
         auto const self = static_cast<Impl*>(data);
 
-        if (self->outputs.find(output) != self->outputs.end())
+        auto const inserted = self->outputs.insert(output);
+
+        if (!inserted.second)
         {
             BOOST_THROW_EXCEPTION(std::runtime_error(
                 "Got wl_surface.enter(wl_output@" +
                 std::to_string(wl_proxy_get_id(reinterpret_cast<wl_proxy*>(output))) +
                 ") for an output the surface is already on"));
         }
-
-        self->outputs.insert(output);
     }
 
     static void on_leave(void* data, wl_surface* /*wl_surface*/, wl_output* output)
     {
         auto const self = static_cast<Impl*>(data);
 
-        if (self->outputs.find(output) == self->outputs.end())
+        auto const erased = self->outputs.erase(output);
+
+        if (!erased)
         {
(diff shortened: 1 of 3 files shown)


src/in_process_server.cpp
```diff
@@ -1809,13 +1817,49 @@ class wlcs::Surface::Impl
...
+
+        if (self->outputs.find(output) != self->outputs.end())
+        {
+            BOOST_THROW_EXCEPTION(std::runtime_error(
+                "Got wl_surface.enter(wl_output@" +
+                std::to_string(wl_proxy_get_id(reinterpret_cast<wl_proxy*>(output))) +
+                ") for an output the surface is already on"));
+        }
+
+        self->outputs.insert(output);
```
RAOF: Non-blocking suggestion: you don't need a two-phase `find`/`insert` here. The [url=https://en.cppreference.com/w/cpp/container/set/insert]return value[/url] of `insert` tells you whether an element was inserted (and so whether there was an element there before).

tests/test_surface_events.cpp
```diff
@@ -588,4 +588,52 @@ TEST_F(ClientSurfaceEventsTest, buffer_release)
...
+    using namespace testing;
+
+    wlcs::Client client{the_server()};
+
+    auto surface = client.create_visible_surface(100, 100);
+    client.roundtrip();
+
+    ASSERT_THAT(surface.current_outputs(), Not(IsEmpty())) << "Precondition failed";
+
+    the_server().move_surface_to(surface, -200, 0);
```
RAOF: This assumes that the compositor will allow surfaces to be moved entirely off the visible area; I would hope that Mir, at least, would disallow this move.

Because WLCS is testing window-management code I think we want to allow compositors to use as much of their normal window management code as possible.

RAOF:
Oh, and also:
bors try

tests/test_surface_events.cpp
wmww: This is unfortunate, but I suppose you are right

src/in_process_server.cpp
wmww: Done

AlanGriffiths:
Curiosity:

bors try

AlanGriffiths (approved):
I think all the contentious stuff got punted. What's left can land:

bors r+

PR #153, opened 2020-03-16
CI: Don't fail if Mir fails a test.

RAOF:
Instead, check that Mir's tests don't die with a signal. That should
be rare enough that this picks up more WLCS bugs than Mir bugs!


AlanGriffiths:
LGTM

bors r+

PR #155, opened 2020-04-15
Refresh CI targets

RAOF:
For some reason time keeps passing!


Saviq:
Network issues.

bors r=RAOF

Saviq:
Meh.

bors r=RAOF

RAOF:
Hm. The network error from Fedora 31 is being suspiciously repeatable. But it *is* a network error…

bors r+

PR #156, opened 2020-04-21
Refactor and augment input region tests

wmww:
There's still a lot going on here, but hopefully this is more clear

### DIFF
--- a/tests/surface_input_regions.cpp
+++ b/tests/surface_input_regions.cpp
@@ -243,7 +243,15 @@ struct InputType
     struct Device
     {
         virtual ~Device() = default;
-        virtual void to_position(std::pair<int, int> position) = 0;
+        virtual void drag_or_move_to_position(std::pair<int, int> position) = 0;
+        virtual void down() = 0; ///< does nothing for touches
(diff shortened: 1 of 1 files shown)


RAOF:
CI failure is genuine; FTBFS on 16.04

tests/surface_input_regions.cpp
```diff
@@ -23,40 +23,43 @@
...
 
     std::string name;
+    std::pair<int, int> surface_size;
     std::vector<Element> elements;
 
     void apply_to_surface(wlcs::Client& client, wl_surface* surface) const
     {
+        if (elements.empty())
+            return;
         auto const wl_region = wl_compositor_create_region(client.compositor());
```
RAOF, line 53: While we're here, does this want to be
`auto const wl_region = wlcs::wrap_wl_object(wl_compositor_create_region(client.compositor());`

wmww: That could work, but this use case is simple enough to not really need it

wmww:
Whoops, looks like this one slid through the cracks
bors r+

PR #157, opened 2020-04-22
XdgSurfaceStable: Add some tests for client protocol errors

RAOF:


RAOF:
Mir currently fails these tests.

RAOF:
Fedora Modular *really* likes to network-fail.

bors r=wmww

PR #158, opened 2020-04-22
XdgPopupTest: Add a role-change test exposing a crash in Mir

RAOF:


RAOF:
This test should *really* be covered by #157 or a general cartesian-product test of existing-role → new-role, but this exposes a crash in Mir *now* and so can be useful before we do the general case.

tests/xdg_popup.cpp
```diff
@@ -720,6 +720,44 @@ TEST_P(XdgPopupTest, popup_configure_is_valid)
     EXPECT_THAT(manager->state.value().height, Gt(0));
 }
 
+/* TODO: This should actually be a parmetrised test handling the cartesian product
+ * of current -> new role changes. For now, this exposes a crash in Mir, so just use this
+ */
+TEST_F(XdgPopupTest, changing_role_is_forbidden)
```
wmww, line 726: I'm not sure this very specific test needs to land in WLCS. It seems to be trying to reproduce a crash in Mir *after* the compositor has already failed to send a protocol error. This test will be useful while debugging the problem, but I don't think it needs to land.

wmww:
Fix in Mir for the use-after-free (still fails as it does not send the correct protocol error): https://github.com/MirServer/mir/pull/1439

tests/xdg_popup.cpp
RAOF: I concur, this was mainly proposed to make it easy to find.

PR #159, opened 2020-04-22
clang-tidy: use const-ref in range-for, rather than copy

RAOF:


RAOF:
Network error in CI. Hopefully transitory?

bors r=wmww

PR #160, opened 2020-05-08
Test that XDG shell surfaces are initially configured correctly

wmww:
Tests https://github.com/MirServer/mir/issues/936 and similar situations


wmww:
You are correct

PR #161, opened 2020-05-12
Add WlProxy

wmww:
This is the result of my talk on IRC with @RAOF last night. It will...
- Add a new `WlProxy<T>` type
- Rename `acquire_interface()` to `bind_if_supported()`
- ~Make `bind_if_supported()` a template that returns a `WlProxy`~
- Move the ownership of some globals out of the `Client` and into individual tests
- Refactor `bind_if_supported()` to not create a temporary registry


wmww:
Damn GCC bug https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56480

wmww:
Ah! I've reproduced the clang-specific crash

wmww:
Closing in favor of 162

PR #165, opened 2020-08-10
Test pointer leaves surface during move and resize

wmww:
Adds a test for XDG surfaces interactive resize, and tests that the cursor leaves the surface during move and resize. The latter currently fails on Mir for both cases and for both v6 and stable.

### DIFF
--- a/tests/xdg_toplevel_stable.cpp
+++ b/tests/xdg_toplevel_stable.cpp
@@ -164,7 +164,7 @@ TEST_F(XdgToplevelStableTest, touch_respects_window_geom_offset)
 
 // TODO: set_window_geometry window size (something will need to be added to wlcs)
 
-TEST_F(XdgToplevelStableTest, interactive_move)
+TEST_F(XdgToplevelStableTest, surface_can_be_moved_interactively)
 {
     int window_x = 100, window_y = 100;
     int window_width = 420, window_height = 390;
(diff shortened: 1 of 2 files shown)


tests/xdg_toplevel_stable.cpp
```diff
@@ -206,9 +206,99 @@ TEST_F(XdgToplevelStableTest, interactive_move)
...
+
+    pointer.move_to(start_x, start_y);
+    pointer.left_button_down();
+
+    client.dispatch_until([&](){
+            return button_down;
+        });
+
+    xdg_toplevel_move(toplevel, client.seat(), last_serial);
     client.dispatch_until([&](){
```
RAOF, line 256: I think this would be clearer if you used `add_pointer_leave_notification()` to detect when the pointer leaves, rather than using the button-up event as a proxy for the pointer leave.

tests/xdg_toplevel_stable.cpp
```diff
@@ -222,6 +312,46 @@ TEST_F(XdgToplevelStableTest, interactive_move)
...
+    pointer.move_to(start_x, start_y);
+    pointer.left_button_down();
+
+    client.dispatch_until([&](){
+            return button_down;
+        });
+
+    xdg_toplevel_resize(toplevel, client.seat(), last_serial, XDG_TOPLEVEL_RESIZE_EDGE_TOP_LEFT);
+    client.dispatch_until([&](){
+            return !button_down;
```
RAOF, line 351: Likewise, it'd be nice to have the “pointer leaves surface” test gated on the pointer_leave event, rather than the button state.

wmww:
I switched to checking `client.window_under_cursor()`.

PR #166, opened 2020-08-11
Update spread images

wmww:
And add Fedora 32

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,8 +44,8 @@ env:
   - SYSTEM=ubuntu-19.10 VARIANT=gcc
   - SYSTEM=ubuntu-devel VARIANT=gcc
   - SYSTEM=ubuntu-devel VARIANT=clang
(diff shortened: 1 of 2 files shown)


wmww:
@AlanGriffiths I'm going to add Debian sid in another PR, so we can take our time reviewing without CI being broken while we do it.

.travis.yml
```diff
@@ -39,13 +39,12 @@ env:
   - DEBFULLNAME="Mir CI Bot"
   - DEBEMAIL="mir-ci-bot@canonical.com"
   jobs:
-  - SYSTEM=ubuntu-16.04 VARIANT=gcc
   - SYSTEM=ubuntu-18.04 VARIANT=gcc
-  - SYSTEM=ubuntu-19.10 VARIANT=gcc
+  - SYSTEM=ubuntu-20.10 VARIANT=gcc
```
AlanGriffiths, line 43: I think you mean 20.04

wmww: Yep

PR #167, opened 2020-08-11
Add Debian sid to CI

wmww:
Similar to https://github.com/MirServer/mir/pull/1662, stacked on #166

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -9,6 +9,8 @@ backends:
             - ubuntu-20.04
             - ubuntu-devel:
                 image: ubuntu-daily:devel/amd64
+            - debian-sid:
+                image: debian-cloud-testing/debian-sid
             - fedora-31
             - fedora-32
             - alpine-3.11
--- a/spread/build/debian/task.yaml
+++ b/spread/build/debian/task.yaml
@@ -1,0 +1,34 @@
+summary: Build (on Debian)
+systems: [debian-*]
+
+execute: |
+    # Grab builds of Mir git master
+    add-apt-repository ppa:mir-team/dev
+
+    apt-get update
+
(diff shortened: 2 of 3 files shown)


wmww:
@Saviq I don't really know what I'm doing here, you may be helpful

spread.yaml
```diff
@@ -9,6 +9,8 @@ backends:
             - ubuntu-20.04
             - ubuntu-devel:
                 image: ubuntu-daily:devel/amd64
+            - debian-sid:
+                image: debian-cloud-testing/debian-sid
```
RAOF, line 13: We use LXD for wlcs CI, which has a different set of images to GCE.

`lxc image list images:debian` lists a bunch of images; we probably want `debian/sid`, and the way the LXD backend of spread works means that the system name `debian-sid` will find the right image without being manually specified.

spread/build/debian/task.yaml
```diff
@@ -0,0 +1,34 @@
+summary: Build (on Debian)
+systems: [debian-*]
+
+execute: |
+    # Grab builds of Mir git master
+    add-apt-repository ppa:mir-team/dev
```
RAOF, line 6: *Pretty* sure this will only work properly on Ubuntu. We might just need to build against the latest Mir release in Debian?

AlanGriffiths: That looks to be 1.8: https://buildd.debian.org/status/package.php?p=mir

But why would the wlcs build depend on Mir? (I know we do it on our infrastructure, but it shouldn't be a hard dependency anywhere else.)

AlanGriffiths:
This has been neglected for over 6 months. If anyone cares they can fix it up and resubmit.

PR #170, opened 2020-08-12
Add surface type and input method abstractions

wmww:
This moves the abstractions added in #156 into WLCS proper. This allows parameterized tests to easily test with all the input methods and surface types they desire (instead of the patchwork of flowed approaches we currently employ).

### DIFF
--- a/include/input_method.h
+++ b/include/input_method.h
@@ -1,0 +1,83 @@
+/*
+ * Copyright © 2020 Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: William Wold <william.wold@canonical.com>
+ */
+
+#ifndef WLCS_INPUT_METHOD_H_
+#define WLCS_INPUT_METHOD_H_
+
+#include <string>
+#include <vector>
+#include "in_process_server.h"
+
+namespace wlcs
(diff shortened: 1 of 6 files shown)


src/surface_builder.cpp
```diff
@@ -0,0 +1,118 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: William Wold <william.wold@canonical.com>
+ */
+
+#include "surface_builder.h"
+
+auto wlcs::SurfaceBuilder::all_surface_types() -> std::vector<std::shared_ptr<SurfaceBuilder>>
+{
+    // All this pointer casting nonsense is only to make 16.04 GCC happy
```
RAOF, line 23: Maybe we could drop 16.04? :wink:

include/input_method.h
```diff
@@ -0,0 +1,81 @@
...
+    struct Device
+    {
+        virtual ~Device() = default;
+        /// 1. Unclicks button or raises touch if currently "down"
+        /// 2. Moves to new positon
+        /// 3. Touches/clicks down at that new position
+        virtual void move_to(std::pair<int, int> position) = 0;
+        /// MUST be called in the "down" state (can not be the first call)
+        /// Moves device to new position while remaining in the "down" state
+        virtual void drag_to(std::pair<int, int> position) = 0;
```
RAOF, line 43: This seems kinda a strange API?

The common API between pointer and touch is really the touch API - `down_at`, `move`, and `up` can all be easily implemented for both, and can implement `move_to` and `drag_to` as specified here.

AlanGriffiths: > The common API between pointer and touch is really the touch API

That shouldn't be the case. Touch and pointer are different abstractions and any commonality is accidental.
Touch can have sizes and multiple touches, pointers can't.
Pointers have multiple buttons, and wheels touches don't.

src/surface_builder.cpp
AlanGriffiths: > Maybe we could drop 16.04? wink

+1

wmww: Oh, wait, I already did drop 16.04.

include/input_method.h
wmww: I agree with both. The common API is not conceptually the touch API, but it is very similar to what the WLCS touch API currently is.

PR #173, opened 2020-08-25
Test version 2 and 3 features of zwlr_layer_shell_v1

wmww:
Note that the major unstable version of layer shell is still 1, so all the changes are backwards compatible. The important change is the addition of a `.set_layer()` request. Tests are skipped on compositors that don't support a new enough layer shell.


RAOF (commented):
This isn't testing the v3 features, is it?

(It'd also be nice to test that `set_layer` is treated as double-buffered state, but that's just an improvement.

wmww:
Version 3 seems to have just added a `.destroy()`, which was also already being auto-generated. Not sure exactly what the difference is, but didn't think it was important to test.

RAOF (changes requested):
Hm, now that I think of it, I think you'll need to manually specialise the `WlInterfaceDescriptor<wlr_layer_shell_v1>` struct, like the `wl_output` one in `in_process_server.h:56-75`. You need to conditionally send the `.destroy()` request, otherwise you'll send an unrecognised request to a v1-supporting server.

RAOF:
(You can easily test the `.destroy()` method by using `client.bind_if_supported<zwlr_layer_shell_v1>(wlcs::AtLeastVersion{3});`

wmww:
`destroy_request_not_sent_when_not_supported` passes with the interface descriptor overload and fails without it, so I think I did it right.

AlanGriffiths (approved):
LGTM

@RAOF I think your comments have been addressed?

RAOF (approved):
Yes, that looks good now. (Although it looks like maybe there are now merge conflicts?)

RAOF:
bors delegate=wmww

AlanGriffiths:
> there are now merge conflicts

Trivial ones. Fixed

PR #174, opened 2020-08-27
Fix SurfaceBuilder/InputMethod stream operators

wmww:
To work with gtest, these operators need to be in `std`

### DIFF
--- a/include/input_method.h
+++ b/include/input_method.h
@@ -75,6 +75,9 @@ struct TouchInputMethod : InputMethod
 
 }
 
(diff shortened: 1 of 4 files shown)


RAOF (approved):
Oh! I initially thought “hm, I thought GTest required `operator<<(…, Foo)` to be in the same namespace as `Foo` and thought this was wrong. But we're printing `std::shared_ptr<Foo>`, so `std` *is* the correct namespace.

bors r+

PR #175, opened 2020-08-27
Destroy surface roles before surfaces

wmww:
Previously in `Client::create_xdg_shell_stable_surface()` and similar methods, we were using `Client::run_on_destruction()` to clean up the surface role objects. Issue with that is it is technically illegal to destroy a surface before destroying it's role (we should probably test that this errors and make Mir send an error, but that's another issue altogether). In this PR I allow adding destruction callbacks to surfaces instead of just clients. They are run before the surface is destroyed.

### DIFF
--- a/src/in_process_server.cpp
+++ b/src/in_process_server.cpp
@@ -1668,6 +1668,10 @@ public:
             }
         }
 
+        for (auto const& callback: destruction_callbacks)
+            callback();
+        destruction_callbacks.clear();
+
         wl_surface_destroy(surface_);
     }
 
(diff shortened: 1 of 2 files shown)


src/in_process_server.cpp
```diff
@@ -734,7 +734,7 @@ class wlcs::Client::Impl
         auto xdg_surface = std::make_shared<XdgSurfaceV6>(client, surface);
         auto xdg_toplevel = std::make_shared<XdgToplevelV6>(*xdg_surface);
 
-        run_on_destruction([xdg_surface, xdg_toplevel]() mutable
+        surface.run_on_destruction([xdg_surface, xdg_toplevel]() mutable
```
RAOF, line 737: This would really be clearer with `FnOnce`, wouldn't it. We can't *actually* make these `unique_ptr`s, because we can't get the compiler to move-only the lambda?

RAOF: I guess it might be a *little* clearer if you moved the `shared_ptr`s into the lambda capture, but I'm not going to block for that.

RAOF (approved):
The thought occurs that `run_before_destruction` might be a more descriptive name, but this is consistent with the existing code.

bors r+

PR #176, opened 2020-08-28
Use XDG surface with geometry

wmww:
Allow the XDG surface builder to set the window geometry and add a surface with geometry set to `all_surface_types`


RAOF (approved):
Yup, looks sensible enough.

bors r+

PR #177, opened 2020-08-28
Update touch tests

wmww:
This ports the touch tests to use the new surface builder abstraction. It also adds a new test, `sends_touch_up_on_surface_destroy`. This currently fails on Mir.


RAOF (approved):
It's not entirely obvious that the “send touch up on surface destroy” is mandated by the protocol, but it's an obviously-sensible behaviour. Let's enforce it (at least until I've got Weston running against WLCS, and we can see whether Weston agrees :wink: )

bors r+

PR #178, opened 2020-08-28
pointer-confinement

AlanGriffiths:
Adds test infrastructure for pointer_constraints and relative_pointer

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -141,6 +141,7 @@ set(
   include/wl_handle.h
   include/in_process_server.h
   include/primary_selection.h
+  include/relative_pointer_unstable_v1.h
   include/xdg_shell_v6.h
   include/xdg_shell_stable.h
   include/xdg_output_v1.h
@@ -160,6 +161,7 @@ set(
   src/main.cpp
   src/primary_selection.cpp
   src/shared_library.cpp
+  src/relative_pointer_unstable_v1.cpp
   src/xfail_supporting_test_listener.h
   src/xfail_supporting_test_listener.cpp
   src/termcolor.hpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,6 +124,8 @@ GENERATE_PROTOCOL(xdg-shell)
 GENERATE_PROTOCOL(wlr-layer-shell-unstable-v1)
 GENERATE_PROTOCOL(xdg-output-unstable-v1)
 GENERATE_PROTOCOL(wlr-foreign-toplevel-management-unstable-v1)
+GENERATE_PROTOCOL(pointer-constraints-unstable-v1)
+GENERATE_PROTOCOL(relative-pointer-unstable-v1)
 
 option(WLCS_BUILD_ASAN "Build a test runner with AddressSanitizer annotations" ON)
 option(WLCS_BUILD_TSAN "Build a test runner with ThreadSanitizer annotations" ON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,6 +182,7 @@ set(
   tests/surface_input_regions.cpp
   tests/frame_submission.cpp
   tests/primary_selection.cpp
+  tests/relative_pointer.cpp
   tests/subsurfaces.cpp
   tests/xdg_surface_v6.cpp
   tests/xdg_toplevel_v6.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -140,6 +140,7 @@ set(
   include/helpers.h
   include/wl_handle.h
   include/in_process_server.h
+  include/pointer_constraints_unstable_v1.h
   include/primary_selection.h
   include/relative_pointer_unstable_v1.h
   include/xdg_shell_v6.h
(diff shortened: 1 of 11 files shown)


AlanGriffiths:
The PR only contains a few initial tests, but allows for feedback on the approach. (The Mir spike of these protocols doesn't pass even these example tests yet.)

CMakeLists.txt
```diff
@@ -170,6 +176,7 @@ set(
   ${PROTOCOL_SOURCES}
 
   tests/test_bad_buffer.cpp
+        tests/pointer_constraints.cpp
```
wmww, line 179: Indentation

src/pointer_constraints_unstable_v1.cpp
```diff
@@ -0,0 +1,93 @@
...
+    return manager;
+}
+
+wlcs::ZwpConfinedPointerV1::ZwpConfinedPointerV1(
+    ZwpPointerConstraintsV1& manager,
+    wl_surface* surface,
+    wl_pointer* pointer,
+    wl_region* region,
+    uint32_t lifetime) :
+    relative_pointer{zwp_pointer_constraints_v1_confine_pointer(manager, surface, pointer, region, lifetime)},
```
wmww, line 42: Don't we generally format this like
```
    uint32_t lifetime)
    : relative_pointer...
```
?

src/pointer_constraints_unstable_v1.cpp
```diff
@@ -0,0 +1,93 @@
...
+    };
+
+
+wlcs::ZwpLockedPointerV1::ZwpLockedPointerV1(
+    ZwpPointerConstraintsV1& manager,
+    wl_surface* surface,
+    wl_pointer* pointer,
+    wl_region* region,
+    uint32_t lifetime) :
+    relative_pointer{zwp_pointer_constraints_v1_lock_pointer(manager, surface, pointer, region, lifetime)},
```
wmww, line 72: Formatting again

tests/pointer_constraints.cpp
```diff
@@ -0,0 +1,118 @@
...
+    EXPECT_CALL(locked_ptr, locked()).Times(1);
+
+    client.roundtrip();
+}
+
+TEST_F(PointerConstraints, locked_pointer_does_not_move)
+{
+    auto const initial_cursor_position = client.pointer_position();
+    ZwpLockedPointerV1 locked_ptr{pointer_constraints, nw_surface, pointer, nullptr, ZWP_POINTER_CONSTRAINTS_V1_LIFETIME_ONESHOT};
+    EXPECT_CALL(locked_ptr, locked()).Times(AnyNumber());
```
wmww, line 86: Perhaps add `<< "Precondition failed"` to make it clear this isn't the main thing this test is testing

tests/pointer_constraints.cpp
```diff
@@ -0,0 +1,118 @@
...
+    client.roundtrip();
+}
+
+TEST_F(PointerConstraints, when_cursor_moves_onto_surface_locked_pointer_gets_locked_notification)
+{
+    ZwpLockedPointerV1 locked_ptr{pointer_constraints, se_surface, pointer, nullptr, ZWP_POINTER_CONSTRAINTS_V1_LIFETIME_ONESHOT};
+
+    EXPECT_CALL(locked_ptr, locked()).Times(1);
+
+    cursor.move_to(se_middle_x, se_middle_y);
```
wmww, line 110: This seems like an implementation detail. In fact, an annoying amount of stuff about how this protocol behaves are implementation-dependent. I'm find with the more basic stuff (locking on a focused surface working immediately), but this is likely to actually be different from compositor to compositor. For example, some might require clicking on the surface before locking.

src/pointer_constraints_unstable_v1.cpp
AlanGriffiths: No, we normally format like this PR. What you show is [option 3](https://mir-server.io/doc/cppguide/?showone=Constructor_Initializer_Lists#Constructor_Initializer_Lists).

tests/pointer_constraints.cpp
AlanGriffiths: It isn't going to fail. I should, perhaps, use `NiceMock<>`

tests/pointer_constraints.cpp
AlanGriffiths: True, the trouble is we don't (yet) have ways to express the requirement to the compositor. I.e. `the_server().capture(cursor, on(se_surface))`.

src/pointer_constraints_unstable_v1.cpp
wmww: Oh, I forgot we had multiple acceptable options there

tests/pointer_constraints.cpp
wmww: Fair. Maybe add a click here though? I'm not convinced it's desirable for any compositor to capture the pointer on mouse over. It should at least need to be focused, no?

AlanGriffiths: Interestingly, adding the click caused my local iteration of the prototype implementation to pass the test.

wmww (approved):
LGTM
bors r+

PR #181, opened 2020-09-19
Improve Layer Shell tests

wmww:
Improvements include:
- Always set size when required by the protocol
- Test margins
- Test popups are placed correctly (The XDG popup test already tests a verity of positioners for Layer Shell popups, but this tests a verity of layer surface configurations with the same basic positioner)

A bit of refactoring was required.

### DIFF
--- a/tests/wlr_layer_shell_v1.cpp
+++ b/tests/wlr_layer_shell_v1.cpp
@@ -86,7 +86,7 @@ public:
     int static const default_height = 50;
 };
 
-struct LayerAnchor
+struct LayerSurfaceLayout
 {
     template<typename T>
     struct Sides
@@ -94,10 +94,10 @@ struct LayerAnchor
         T left, right, top, bottom;
     };
 
-    static auto get_all() -> std::vector<LayerAnchor>
+    static auto get_all() -> std::vector<LayerSurfaceLayout>
     {
         bool const range[]{false, true};
-        std::vector<LayerAnchor> result;
+        std::vector<LayerSurfaceLayout> result;
         for (auto left: range)
             for (auto right: range)
                 for (auto top: range)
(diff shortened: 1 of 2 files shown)


tests/wlr_layer_shell_v1.cpp
```diff
@@ -82,119 +90,142 @@ class LayerSurfaceTest: public wlcs::StartedInProcessServer
...
     }
 
-    bool const left, right, top, bottom;
+    Sides<bool> const anchor;
+    Sides<int> const margin;
 };
 
-std::ostream& operator<<(std::ostream& os, const LayerAnchor& anchor)
+// Given a LayerSurfaceLayout, emits the margin sides to use in a call to zwlr_layer_surface_v1_set_margin()
+#define SPLAT_MARGIN(layout) (layout).margin.top, (layout).margin.right, (layout).margin.bottom, (layout).margin.left
```
RAOF, line 217: I am not wild about `SPLAT_MARGIN`; I see what you're doing with it, but I don't think the small decrease in verbosity is worth the surprise of a macro expanding into multiple arguments.

RAOF (approved):
Looks (mostly) good to me; sorry for the delay.

I don't think `SPLAT_MARGIN` is appropriate; I think it's it's optimising for writing slightly fewer characters of code at the expense of clarity.

With that removed,
bors r=wmww

(It would also be nice to test the protocol error when not setting size, but the protocol itself is unclear as to what error to send, so :woman_shrugging: )

wmww:
I don't think `bors r=wmww` meant what you thought it meant. I will remove `SPLAT_MARGIN` and PR the changes tomorrow.

tests/wlr_layer_shell_v1.cpp
wmww: This is the problem with code review, you can't get anything fun by it

RAOF:
Oops. `r=` is not the same as `delegate=` :)

PR #182, opened 2020-10-26
Replace the SPLAT_MARGIN macro

wmww:
@RAOF I know you prefer calling the Wayland request functions directly in tests, but is someone getting confused by this indirection really more likely then someone mixing up the margin order and getting confused by that? I'm particularly worried about this because the margin order is weird and that would be a really annoying bug to track down (this is why I had `SPLAT_MARGIN` to begin with)


RAOF (approved):
This looks like a pretty benign indirection, and being able to name the parameters is indeed sensible.

bors r+

PR #183, opened 2020-10-27
layer-shell: Test protocol errors for unspecified surface size with missing anchors

RAOF:


RAOF:
Because I felt like it while I was here.

AlanGriffiths (approved):
Sure, why not?

bors r+

PR #186, opened 2020-12-17
Pointer confinement

AlanGriffiths:
Updated (but still incomplete) pointer confinement tests.

Notably, there are no tests setting a confinement region.


wmww (approved):
LGTM
bors r+

PR #188, opened 2021-02-12
CI: Update Alpine & Fedora images

z3ntu:
Fedora 31 isn't available anymore according to https://us.images.linuxcontainers.org/

And Alpine Linux 3.13 is available since a month or so https://alpinelinux.org/releases/


AlanGriffiths (approved):
LGTM

bors r+

PR #189, opened 2021-02-18
Set the selection when offering cut & paste

AlanGriffiths:
Fixes: https://github.com/MirServer/mir/issues/1940


wmww (approved):
Oh, so that's what I was missing
bors r+

PR #190, opened 2021-02-18
Add tests for popup done event

wmww:
Adds tests that we get a popup done event when popups are dismissed. `does_not_get_popup_done_event_before_button_press` fails due to https://github.com/MirServer/mir/issues/1818. I made the client remember the last serial it got, so we don't need to wire up pointer listeners when that's all we care about.


wmww:
This is now based on #191

AlanGriffiths (approved):
WFM

bors r+

PR #192, opened 2021-03-25
Check relative pointer doesn't suppress motion events

AlanGriffiths:
Check Cursor movement is not propagating to clients correctly when zwp_relative_pointer_manager_v1 is enabled.

C.f. https://github.com/MirServer/mir/issues/1959


wmww (approved):
LGTM
bors r+

PR #193, opened 2021-04-04
Fix FindGtestGmock.cmake for new cmake

z3ntu:
CMake 3.20 uses the upstream CMake config from GTest if possible which
defines GTEST_LIBRARIES instead of GTEST_LIBRARY.

Fixes the build on Alpine Linux.


wmww:
Can you please include something like your PR comment as a code comment so we know in the future why that is there? Thanks!

z3ntu:
Happy with my comment? :)

wmww (approved):
LGTM
bors r+

PR #195, opened 2021-04-21
Test that frame callbacks are not sent immediately

wmww:
Tests the underlying cause of https://github.com/MirServer/mir/issues/1967. Mir master currently fails the `frame_sent_after_delay_with_no_buffer`. https://github.com/MirServer/mir/pull/2006 passes.

### DIFF
--- a/tests/frame_submission.cpp
+++ b/tests/frame_submission.cpp
@@ -79,3 +79,28 @@ TEST_F(FrameSubmission, post_one_frame_at_a_time)
         EXPECT_THAT(frame_consumed, Eq(true));
     }
 }
+
+TEST_F(FrameSubmission, frame_sent_after_delay)
+{
+    auto const surface = client.create_visible_surface(200, 200);
+    ShmBuffer buffer{client, 200, 200};
+    wl_surface_attach(surface, buffer, 0, 0);
+
+    auto frame_consumed = false;
+    submit_frame(frame_consumed);
+    client.roundtrip();
+    EXPECT_THAT(frame_consumed, Eq(false)) << "frame callback sent immediately";
+    wait_for_frame(frame_consumed);
+    EXPECT_THAT(frame_consumed, Eq(true));
+}
+
+TEST_F(FrameSubmission, frame_sent_after_delay_with_no_buffer)
+{
+    auto const surface = client.create_visible_surface(200, 200);
+    auto frame_consumed = false;
+    submit_frame(frame_consumed);
+    client.roundtrip();
+    EXPECT_THAT(frame_consumed, Eq(false)) << "frame callback sent immediately";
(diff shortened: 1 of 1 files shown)


tests/frame_submission.cpp
```diff
@@ -79,3 +79,28 @@ TEST_F(FrameSubmission, post_one_frame_at_a_time)
...
+TEST_F(FrameSubmission, frame_sent_after_delay)
+{
+    auto const surface = client.create_visible_surface(200, 200);
+    ShmBuffer buffer{client, 200, 200};
+    wl_surface_attach(surface, buffer, 0, 0);
+
+    auto frame_consumed = false;
+    submit_frame(frame_consumed);
+    client.roundtrip();
+    EXPECT_THAT(frame_consumed, Eq(false)) << "frame callback sent immediately";
```
AlanGriffiths, line 92: I was surprised that this passes with current Mir. There's simply not going to be a guarantee that Mir doesn't composite and send a frame event before the roundtrip is over.

And when I tried it it against Mir/master fails:
```plain
$ ../wlcs/cmake-build-debug/wlcs cmake-build-debug/lib/miral_wlcs_integration.so --gtest_filter=FrameSubmission*
Note: Google Test filter = FrameSubmission*
[==========] Running 3 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 3 tests from FrameSubmission
[ RUN      ] FrameSubmission.post_one_frame_at_a_time
[       OK ] FrameSubmission.post_one_frame_at_a_time (23 ms)
[ RUN      ] FrameSubmission.frame_sent_after_delay
/home/alan/CLionProjects/wlcs/tests/frame_submission.cpp:92: Failure
Value of: frame_consumed
Expected: is equal to false
  Actual: true (of type bool)
frame callback sent immediately
[  FAILED  ] FrameSubmission.frame_sent_after_delay (15 ms)
[ RUN      ] FrameSubmission.frame_sent_after_delay_with_no_buffer
/home/alan/CLionProjects/wlcs/tests/frame_submission.cpp:103: Failure
Value of: frame_consumed
Expected: is equal to false
  Actual: true (of type bool)
frame callback sent immediately
[  FAILED  ] FrameSubmission.frame_sent_after_delay_with_no_buffer (13 ms)
[----------] 3 tests from FrameSubmission (51 ms total)

[----------] Global test environment tear-down
[==========] 3 tests from 1 test cases run. (51ms total elapsed)
[  PASSED  ] 1 test
[  FAILED  ] 2 tests failed:
[  FAILED  ] FrameSubmission.frame_sent_after_delay_with_no_buffer
[  FAILED  ] FrameSubmission.frame_sent_after_delay
```

wmww: Hm, I guess there isn't anything here to test then

PR #196, opened 2021-05-03
Fix Intermittent ClientSurfaceEventsTest failing on armhf on LP

AlanGriffiths:
Fixes: https://github.com/MirServer/mir/issues/2019


wmww:
I don't have a good understanding of if this test is good or bad. Perhaps @RAOF?

RAOF (approved):
It's annoying that this is the only test which interrogates `wl_buffer.release`, but it *does* make not-necessarily-valid assumptions about compositor behaviour.

We should extract a test for the buffer-attached-but-replaced-before-`wl_surface.commit`, as *that* should be able to be written without making unreasonable compositor assumptions, but that can be a separate PR.

bors r+

PR #197, opened 2021-05-03
[spread] allow `ubuntu-devel` and `fedora-33` to fail for now

Saviq:
It's not open yet / Fedora doesn't get an IP.


AlanGriffiths (approved):
WFM

bors merge

PR #199, opened 2021-05-25
Destroy subsurfaces

wmww:
Seems like a simple oversight


RAOF (approved):
Yup, seems sensible.

bors r+

PR #203, opened 2021-06-19
Test layer shell keyboard interactivity

wmww:
Bumps to the new layer shell version and tests setting keyboard interactivity.

Also changes the layer shell popup tests to use the new on-demend keyboard interactivity, which will be needed for them to behave normally with the incoming Mir updates.


RAOF (approved):
Yup, looks sensible.

bors r+

Issue #205, opened 2021-07-08
Document build instructions and required versions for dependencies ?


vberger:
Hi,

I've been willing to toy around with trying to use WLCS for [Smithay](https://github.com/Smithay/smithay), but I fail to compile WLCS on my dev computer, so I suspect I must be doing something wrong, possibly regarding versions of dependencies.

My attempt was:

```
git clone https://github.com/MirServer/wlcs.git
cd wlcs
mkdir build
cd build
cmake ..
make
```

And compilation fails with the following errors:

```
/home/victor/dev/wlcs/tests/pointer_constraints.cpp:47:5: error: reference to ‘Pointer’ is ambiguous
   47 |     Pointer cursor = the_server().create_pointer();
      |     ^~~~~~~
In file included from /home/victor/dev/wlcs/tests/pointer_constraints.cpp:21:
/home/victor/dev/wlcs/include/in_process_server.h:77:7: note: candidates are: ‘class wlcs::Pointer’
   77 | class Pointer
      |       ^~~~~~~
In file included from /usr/include/gmock/gmock-spec-builders.h:75,
                 from /usr/include/gmock/gmock-function-mocker.h:42,
                 from /usr/include/gmock/gmock.h:61,
                 from /home/victor/dev/wlcs/include/pointer_constraints_unstable_v1.h:26,
                 from /home/victor/dev/wlcs/tests/pointer_constraints.cpp:19:
/usr/include/gmock/gmock-matchers.h:4827:47: note:                 ‘template<class InnerMatcher> testing::internal::PointerMatcher<InnerMatcher> testing::no_adl::Pointer(const InnerMatcher&)’
 4827 | inline internal::PointerMatcher<InnerMatcher> Pointer(
      |                                               ^~~~~~~
```

For completeness, the cmake output with the versions of compilers & libraries it found:

```
-- The C compiler identification is GNU 11.1.0
-- The CXX compiler identification is GNU 11.1.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found PkgConfig: /usr/bin/pkg-config (found version "1.7.3")
[...]

PR #206, opened 2021-07-08
cmake: always set GTEST_LIBRARY if it isn't

vberger:
Previously, the GTEST_LIBRARY would only be set manually set when
GMock is not built from source, causing likage failures on systems
where Gmock is built from source and system-provided GTest does not
set it.

cc #205


wmww (approved):
LGTM
bors r+

PR #207, opened 2021-07-08
Fix build when using GTest-1.11

vberger:
Some change in GTest 1.11 headers caused a conflict on the name
`Pointer`. Removing the unconditional import of namespace `testing`
in favor of individual imports of the used items fixes it.

cc #205

### DIFF
--- a/tests/pointer_constraints.cpp
+++ b/tests/pointer_constraints.cpp
@@ -24,7 +24,11 @@
 
 #include <memory>
 
-using namespace testing;
+using testing::AnyNumber;
+using testing::Eq;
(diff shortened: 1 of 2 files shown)


RAOF (approved):
Looks good to me.

What distro are you using that has GTest 1.11? If it's easy to add that to our testing matrix I might do that before landing this; if it's not easy then we can just land this as-is.

vberger:
It was on my dev laptop, that's on Archlinux.

wmww (approved):
Seems to fix the build on my Arch box
bors r+

PR #209, opened 2021-08-10
Test text input v3 and input method v2

wmww:
Tests for https://github.com/MirServer/mir/pull/2139

### DIFF
--- a/src/protocol/input-method-unstable-v2.xml
+++ b/src/protocol/input-method-unstable-v2.xml
@@ -2,6 +2,9 @@
 <protocol name="input_method_unstable_v2">
 
   <copyright>
+    Copyright © 2008-2011 Kristian Høgsberg
+    Copyright © 2010-2011 Intel Corporation
+    Copyright © 2012-2013 Collabora, Ltd.
     Copyright © 2012, 2013 Intel Corporation
     Copyright © 2015, 2016 Jan Arne Petersen
     Copyright © 2017, 2018 Red Hat, Inc.
@@ -29,14 +32,14 @@
 
   <description summary="Protocol for creating input methods">
     This protocol allows applications to act as input methods for compositors.
-    
+
     An input method context is used to manage the state of the input method.
 
     Text strings are UTF-8 encoded, their indices and lengths are in bytes.
 
     This document adheres to the RFC 2119 when using words like "must",
     "should", "may", etc.
-    
+
     Warning! The protocol described in this file is experimental and
     backward incompatible changes may be made. Backward compatible changes
     may be added together with the corresponding interface version bump.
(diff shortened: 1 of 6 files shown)


wmww:
More tests can always be added but now there's enough to be useful

src/protocol/input-method-unstable-v2.xml
```diff
@@ -0,0 +1,404 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="input_method_unstable_v2">
```
RAOF, line 2: Where does this protocol definition come from? As far as I can tell, `wayland-protocols` has only v1 and doesn't even have a PR for adding v2?

We probably want to have a canonical source for this :)

RAOF: Could *we* drive a `wayland-protocols` PR?

AlanGriffiths: There's some discussion of this protocol version in the Mir PR https://github.com/MirServer/mir/pull/2139#issuecomment-897971091

And, yes, we ought to be able to drive a `wayland-protocols` PR.

wmww: Whoops, these are the still the old versions that Squeekboard had. The first step is making sure [Squeekboard matches us and wlroots](https://gitlab.gnome.org/World/Phosh/squeekboard/-/merge_requests/472). After that we can talk to both of those teams to see what standard place these belong in (wayland-protocols or wlr-protocols). Not even Layer Shell is in wayland-protocols (it's been in [PR](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/28) for over a year, so I'm not super enthused about trying to get something in it.

RAOF: I don't think we need to care *too* much about *landing* a PR in wayland-protocols (although that would obviously be ideal). I *do* think we need to care about there being a single, canonical place to get the protocol from. There are already at least two slightly different versions of this protocol around!

Since this is a v2 to a `wp_` protocol already in wayland-protocols, wayland-protocols is the only sensible place to put it.

AlanGriffiths (approved):
I don't think the valid concern about a canonical source for this protocol version should prevent landing. These tests may be minimal, but they are relevant.

wmww:
It should not block, as resolving it may take some time. I've started asking around on IRC to figure out where the protocol came from and where it should be PRed.

src/protocol/input-method-unstable-v2.xml
wmww: There is extensive discussion of the whole ecosystem in [these](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/39) [two](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/40) `wayland-protocols` issues. The link they give for input-method-v2 is to the wlroots source tree (the same place I got it).

tests/text_input_v3_with_input_method_v2.cpp
```diff
@@ -0,0 +1,174 @@
...
+    EXPECT_CALL(text_input, enter(_))
+        .WillOnce(SaveArg<0>(&entered));
+    create_focussed_surface();
+    EXPECT_THAT(entered, Eq(app_surface.value().operator wl_surface*()));
+}
+
+TEST_F(TextInputV3WithInputMethodV2Test, text_input_leaves_surface_on_unfocus)
+{
+    create_focussed_surface();
+    EXPECT_CALL(text_input, enter(_)).Times(0);
```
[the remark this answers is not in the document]
wmww: The text input is only given enter events for surfaces belonging to the client that created it. The other surface belongs to a different client.

tests/text_input_v3_with_input_method_v2.cpp
```diff
@@ -0,0 +1,174 @@
...
+    EXPECT_CALL(text_input, leave(app_surface.value().operator wl_surface*()));
+
+    // Create a 2nd client with a focused surface
+    wlcs::Client other_client{the_server()};
+    auto other_surface = other_client.create_visible_surface(100, 100);
+    the_server().move_surface_to(other_surface, 200, 200);
+    pointer.move_to(210, 210);
+    pointer.left_click();
```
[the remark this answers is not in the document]
wmww: That may be a good idea

AlanGriffiths:
No reason for further delay:

bors merge

PR #210, opened 2021-08-10
WlHandle: move specialization check to handle constructor

wmww:
There are a number of ways one might end up with a `WlHandle` without using `wrap_wl_object()`. This makes sure the specialization check applies to all of them.


RAOF (approved):
Sure.

bors r+

PR #211, opened 2021-08-10
Add method_event_impl()

wmww:
This helper makes it easy and less error prone to implement a Wayland listener using a class method. All that needs to be specified is the pointer-to-member function, and the rest is taken care of by `method_event_impl()`. See an example of it's usage in #209.


RAOF (approved):
Also looks sensible.

bors r+

PR #217, opened 2022-01-06
Test that input method can be deactivated

wmww:
Tests https://github.com/MirServer/mir/issues/2269


tests/text_input_v3_with_input_method_v2.cpp
```diff
@@ -102,6 +102,42 @@ TEST_F(TextInputV3WithInputMethodV2Test, input_method_can_be_enabled)
     input_client.roundtrip();
 }
 
+TEST_F(TextInputV3WithInputMethodV2Test, input_method_can_be_disabled)
+{
+    create_focussed_surface();
+    Expectation a = EXPECT_CALL(input_method, activate());
```
RAOF, line 108: You could probably just use `InSequence` here, right? And then the `deactivate` would also be sequenced after the `activate`/`done`?

RAOF (approved):
Seems reasonable. Feel free to accept the `InSequence` suggestion, or not.

bors delegate=wmww

RAOF (approved):
bors r+

PR #221, opened 2022-02-18
hirsute is obsolete

AlanGriffiths:


Saviq:
I'll have a PR up soon to rework this into the same way Mir itself works.

PR #223, opened 2022-02-18
Test popups are dismissed in the correct order

wmww:
It unfortunately required a fair bit of code churn to be able to make popups that are children of other popups. The actual logic of other popup tests should not be effected.


tests/xdg_popup.cpp
```diff
@@ -732,19 +763,49 @@ TEST_P(XdgPopupTest, grabbed_popup_gets_done_event_when_new_toplevel_created)
...
+    top_popup_manager->client->roundtrip();
+    auto sub_popup_manager = top_popup_manager->child();
+    sub_popup_manager->map_popup(positioner);
+    top_popup_manager->client->roundtrip();
+
+    InSequence seq;
+    EXPECT_CALL(*sub_popup_manager, popup_done());
+    EXPECT_CALL(*top_popup_manager, popup_done());
+
+    // Create a new toplevel, which will dismiss the popups
```
RAOF, line 807: As far as I can tell this is *not* guaranteed by the protocol.

Indeed, I don't think that the behaviour this test expects is guaranteed by the protocol:

> When compositors choose to dismiss a popup, they *may* dismiss every
> nested grabbing popup as well. When a compositor dismisses popups, it
> will follow the same dismissing order as required from the client.
(Emphasis mine).

I don't think that the compositor is required to dismiss any but the topmost popup, *but* if it dismisses any others it must do so in the specified order. (This seems like maybe a spec bug, and I'll file a couple once I recover(?!) my Freedesktop login).

I *think* it's guaranteed that clicking somewhere outside the surface will dismiss the popup, so maybe the test should be doing that?

tests/xdg_popup.cpp
```diff
@@ -174,41 +174,42 @@ class XdgPopupManagerBase
...
     void unmap_popup()
     {
         clear_popup();
-        popup_surface = std::experimental::nullopt;
+        popup_surface = std::nullopt;
     }
 
-    virtual auto popup_position() const -> std::experimental::optional<std::pair<int, int>> = 0;
+    virtual auto popup_position() const -> std::optional<std::pair<int, int>> = 0;
+    virtual auto child() -> std::unique_ptr<XdgPopupManagerBase> = 0;
```
RAOF, line 201: I was confused by this on the first-pass review. This isn't *returning* whatever child the popup has, right? It is creating a new child popup?

If it's going to do that it should have some sort of verb in the name; `create_child_popup()` maybe?

RAOF (changes requested):

RAOF:
Also, I seem to have broken the build by removing `#include <experimental/optional>`, which I'll fix.

tests/xdg_popup.cpp
AlanGriffiths: I think it poorly worded, but I'm not sure your interpretation is a reasonable behaviour.

I think it a spec bug and that the intent is "they _must_ dismiss every nested grabbing popup as well"

Did you log a bug?

wmww:
It seems at some point Mir stopped sending `.popup_done` events on click outside of the popup. Will investigate.

wmww:
After testing with other compositors I think Mir is doing the right thing. Test should now work.

RAOF (approved):
LGTM (with the fix to not `#include <experimental/optional>`, which I've committed on top)

PR #231, opened 2022-04-05
[ppa] fix LP credentials and PPAs

### LINKED ISSUE #229: PPA upload not working
https://github.com/MirServer/wlcs/actions/workflows/ppa-upload.yml

Saviq:
Fixes #229 

---

There isn't a `mir-team/wlcs` PPA, but do we even want it?

### DIFF
--- a/tools/ppa-upload.sh
+++ b/tools/ppa-upload.sh
@@ -2,6 +2,12 @@
 
 set -e
 
+LP_CREDS="$1"
+if [ -z "${LP_CREDS}" ]; then
+  echo "ERROR: pass the Launchpad credentials file as argument" >&2
+  exit 2
+fi
+
 if [ -z "${RELEASE}" ]; then
   echo "ERROR: RELEASE environment variable needs to be set to the" >&2
   echo "  target Ubuntu version." >&2
@@ -86,7 +92,7 @@ try:
     authorization_engine=RequestTokenAuthorizationEngine("production",
                                                          "mir-ci"),
     credential_store=UnencryptedFileCredentialStore(
-      os.path.expanduser("~/.launchpadlib/credentials")
+      os.path.expanduser("${LP_CREDS}")
     )
   )
 except NotImplementedError:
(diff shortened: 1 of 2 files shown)


tools/ppa-upload.sh
```diff
@@ -97,7 +103,7 @@ series = ubuntu.getSeries(name_or_version=os.environ['RELEASE'])
 
 mir_team = lp.people["mir-team"]
 rc_ppa = mir_team.getPPAByName(name="wlcs-rc")
-release_ppa = mir_team.getPPAByName(name="wlcs")
+release_ppa = mir_team.getPPAByName(name="release")
```
RAOF, line 106: Hm. This is going to cause WLCS releases to immediately hit the Mir release PPA, which was not our plan. The plan was to manually copy from a WLCS release PPA (which it seems I neglected to actually set up!) into Mir release PPA when we were ready, as new WLCS releases can result in Mir CI failing.

That said, new WLCS releases aren't exactly a daily occurrence *and* we now have very good tools for marking WLCS tests as expected-fail, so maybe direct-to-Mir-release *is* the right choice now.

Saviq: Yeah I believe we need to re-think this a bit. Do we need separate `wlcs-` PPAs? Don't we hide things by not running Mir `-dev` with WLCS `dev` until such a time they meet in the Mir release PPA?

RAOF: The idea was that if we push WLCS builds straight into Mir CI then WLCS can break Mir CI without any code changes in Mir, which is a bit annoying.

That said, it would be easy to notice breakage and, as I said, we now have good tools for `XFAIL`ing tests, so maybe we *should* be breaking Mir CI early?

RAOF: Which is to say: @AlanGriffiths, @wmww, @graysonguarino: opinions?

Saviq: I'd 👍 just using `mir-team/dev`, `/rc`, `/release` PPAs for WLCS.

AlanGriffiths: I don't think there's an approach without drawbacks. Let's just try it, if it breaks we will fix.

RAOF (approved):
With that roaring sound of approval,

bors r+

PR #235, opened 2022-06-08
Do not bind to interface versions newer than supported

### LINKED ISSUE #234: WLCS will happily bind a global version higher than the one defined in the wayland.xml it was compiled for
We stumbled on this by chance when working on Smithay's test suite.

In our case, the wayland.xml used to compile Smithay and the one used by WLCS are apparently different (I guess WLCS uses the system one), and while Smithay advertises wl_output version 4, the wayland.xml WLCS was compiled against only has version 3.
Yet, WLCS will bind version 4 of the global, and then libwayland chockes when it receives a `wl_output.name` event which only exists since version 4.

wmww:
Fixes #234

### DIFF
--- a/include/version_specifier.h
+++ b/include/version_specifier.h
@@ -20,7 +20,8 @@
 #define WLCS_VERSION_SPECIFIER_H_
 
 #include <cstdint>
-#include <experimental/optional>
+#include <optional>
+#include <string>
 
 namespace wlcs
 {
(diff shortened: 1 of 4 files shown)


include/version_specifier.h
```diff
@@ -20,7 +20,8 @@
 #define WLCS_VERSION_SPECIFIER_H_
 
 #include <cstdint>
-#include <experimental/optional>
+#include <optional>
+#include <string>
```
wmww, line 24: Unclear why the string include is now required to compile, but it is. Perhaps experimental/optional included it but optional doesn't?

RAOF: Well, we absolutely use `std::string` in the interfaces, so we should definitely be `#include`ing it.

RAOF (approved):
Huh. Yeah, that's sensible.

bors r+

Saviq:
Networking…

bors retry

PR #238, opened 2022-06-20
Make zxdg_shell_v6 ExpectedlyNotSupported if not in supported_extensions

### LINKED ISSUE #237: xdg_shell_unstable_v6 is currently not optional in the test suite
I'm currently trying to expand the set of WLCS tests we run on Smithay, and it appears that many tests involving `xdg_shell_unstable_v6` will fail if the global is missing, even if the global is not listed as a `supported_extensions`.

For test groups like `XdgSurfaceV6Test.*`, it easy to handle: we just don't run the group.

But several other groups contain V6 tests, for example `AnchorRect/XdgPopupPositionerTest.*` contains some `xdg_shell_unstable_v6_popup_placed_correctly/*` test, which currently fail on Smithay because we've dropped `xdg_shell_unstable_v6` support.

It would be very convenient if those tests could be skipped instead if `zxdg_shell_v6` is not listed as part of `supported_extensions`.

I can try to make a PR for this if you don't have time for it, but I'll need some guidance on how to do it properly.

vberger:
I have tested that it does indeed causes GTest to skip the related tests.

Fixes #237

### DIFF
--- a/src/in_process_server.cpp
+++ b/src/in_process_server.cpp
@@ -809,7 +809,21 @@ public:
 
     zxdg_shell_v6* the_xdg_shell_v6() const
     {
-        return xdg_shell_v6;
+        if (xdg_shell_v6)
+        {
+            return xdg_shell_v6;
+        }
+        else
+        {
+            if (!supported_extensions || !supported_extensions->count("zxdg_shell_v6"))
+            {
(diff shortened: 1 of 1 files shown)


AlanGriffiths (approved):
bors merge

AlanGriffiths:
Thanks!

PR #239, opened 2022-06-20
Destroy xdg_toplevel before xdg_surface on cleanup

vberger:
This is mandated by the protocol.


RAOF (approved):
Correct!

(Looks like the Fedora CI failure should be a transient infrastructure failure)

bors r+

PR #245, opened 2022-10-20
Create surface in primary selection tests

wmww:
Mir's primary selection implementation will not allow clients to paste when they do not have focus, so the tests fail without this.


Saviq:
NB: the devel failures are likely fallout from release and pending switchover, but won't block merging.

RAOF (approved):
We should generally be more explicit with our focus requirements in WLCS tests :grin: 

This is sensible.

bors r+

Saviq:
bors retry

PR #252, opened 2022-11-21
spread: use spread-mir-ci

Saviq:
We're maintaining a spread fork there.

---

Rather than #250. The snap has MirServer/spread-mir-ci#2, fixing the issue.

### DIFF
--- a/.github/workflows/spread.yml
+++ b/.github/workflows/spread.yml
@@ -45,7 +45,7 @@ jobs:
     - name: Set up Spread
       run: |
         set -euo pipefail 
-        sudo snap install spread-mir-ci --beta
+        sudo snap install spread-mir-ci
         sudo snap run lxd init --auto
 
     - name: Check out code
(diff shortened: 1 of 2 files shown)


Saviq:
@RAOF after https://github.com/MirServer/spread-mir-ci/commit/f34932137a2cdb33dd046cf5e2cc87c445a08f2e this is green :)

.github/workflows/spread.yml
```diff
@@ -45,15 +45,13 @@ jobs:
     - name: Set up Spread
       run: |
         set -euo pipefail 
-        curl -s -O https://people.canonical.com/~chrishr/spread.snap
+        sudo snap install spread-mir-ci
         sudo snap run lxd init --auto
```
RAOF, line 49: There seem to be some GitHub actions we could use instead of setting up LXD ourselves. Maybe something to look into later.

RAOF (approved):
:tada:

bors r+

PR #253, opened 2022-12-05
Fix BadBufferTest.test_truncated_shm_file

RAOF:
The protocol error should be generated on the `wl_shm_pool`,
as that's the object that has been asked for an invalid operation.

Additionally, fix the assumption that `wl_display_get_protocol_error`
will fill in the `interface` parameter. In the case where the server
sends an error on an object the client has already deleted, the
`interface` parameter will be set to null.

### DIFF
--- a/include/in_process_server.h
+++ b/include/in_process_server.h
@@ -299,35 +299,12 @@ private:
 class ProtocolError : public std::system_error
 {
 public:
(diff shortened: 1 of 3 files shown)


AlanGriffiths (approved):
Makes sense to me

bors merge

PR #257, opened 2022-12-12
test

AlanGriffiths:
https://github.com/MirServer/mir/issues/2748 just might be a teardown synchronization problem. This adds a bit of sync that we see in some other tests

### DIFF
--- a/tests/wlr_layer_shell_v1.cpp
+++ b/tests/wlr_layer_shell_v1.cpp
@@ -41,6 +41,11 @@ public:
     {
     }
 
+    ~LayerSurfaceTest()
+    {
+        client.roundtrip();
+    }
+
     void commit_and_wait_for_configure()
     {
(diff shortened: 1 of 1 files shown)


tests/wlr_layer_shell_v1.cpp
```diff
@@ -41,6 +41,11 @@ class LayerSurfaceTest: public wlcs::StartedInProcessServer
     {
     }
 
+    ~LayerSurfaceTest()
+    {
+        client.roundtrip();
```
RAOF, line 46: I don't see how this can have any effect?

The client only ever gets callbacks from the server when explicitly dispatched; this adds an explicit dispatch point, but without this explicit dispatch all that will be happening in `~LayerSurfaceTest` is ~LayerSurfaceV1()`, `~Surface()`, then `~Client()`, none of which perform dispatch?

PR #260, opened 2023-01-13
pkgconfig: Use better path concatenation

OPNA2608:
A correction to #258, based on recommendations from [jtojnar/cmake-snips#concatenating-paths-when-building-pkg-config-files](https://github.com/jtojnar/cmake-snips/tree/a9d72829ab0cb42e541c9975e698d6b8282610ed#concatenating-paths-when-building-pkg-config-files).

Unconditionally using `CMAKE_INSTALL_FULL_*DIR`s in a pkg-config file is slightly wrong, because it is expected that the variables are made up of other variables, all of which can be overridden by the user.

CMake added a `cmake_path` function which can do the required concatenation without the added function, but it's only available on version 3.20 while the current CMakeLists specifies a minimum version of 3.5.

I've checked that this gives regular `${prefix}/<dir>` entries again when the `CMAKE_INSTALL_*DIR`s are relative, and the `CMAKE_INSTALL_*DIR`s verbatim when they are absolute.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -281,6 +281,10 @@ install(
   DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
 )
 
+include(JoinPaths)
+join_paths(PKGCONFIG_BINDIR "\${prefix}" "${CMAKE_INSTALL_BINDIR}")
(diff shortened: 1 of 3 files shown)


RAOF (approved):
Yeah, ok.

bors r+

Issue #268, opened 2023-02-01
Fix scrolling tests since addition of value120


graysonguarino:
This issue is dependent on #267 landing. Since the smooth scrolling update to `wl_pointer`, this seems to break tests such as `VirtualPointerV1Test.when_virutal_pointer_scrolls_with_steps_client_sees_axis_descrete`.

AlanGriffiths:
The test is wrong: Clients on V8+ should be getting value120 notifications. (We should test V7- and V8+ - but not sure that wlcs supports that yet)

AlanGriffiths:
There's more to this than I thought:

wlcs is still binding `wl_seat` to v6 - so it should be getting the `axis_discrete` events and be unaffected by anything from v8 (specifically `axis_value120`)

That suggests there is an issue in Mir.

AlanGriffiths:
From the recent duplicate:

```
Uninteresting mock function call - returning directly.
    Function call: axis_value120(1, 480)
./tests/wlr_virtual_pointer_v1.cpp:241: Failure
Actual function call count doesn't match EXPECT_CALL(listener, axis_discrete(WL_POINTER_AXIS_HORIZONTAL_SCROLL, 4))...
         Expected: to be called once
           Actual: never called - unsatisfied and active
```

For some reason, Mir is sending `axis_value120` events. It would be good to confirm which version of `wl_seat` was bound in this session - because when I try to reproduce everything works fine.

AlanGriffiths:
Even more curious: if I hack the version check in Mir to force `axis_value120` events to be sent, then the failure mode is:

```
[ RUN      ] VirtualPointerV1Test.when_virtual_pointer_scrolls_with_steps_client_sees_axis_descrete
Checking version_supports_axis_value120() bind version for wl_seat is 6Checking version_supports_axis_value120() bind version for wl_seat is 6wl_display@1: error 3: Mir tried to send wl_pointer@13.axis_value120 to object with version 6 (requires version 8)
unknown file: Failure
C++ exception with description "Wayland protocol error: 3 on interface wl_display v1" thrown in the test body.

./tests/wlr_virtual_pointer_v1.cpp:256: Failure
Actual function call count doesn't match EXPECT_CALL(listener, axis_discrete(WL_POINTER_AXIS_HORIZONTAL_SCROLL, 4))...
         Expected: to be called once
           Actual: never called - unsatisfied and active

./tests/wlr_virtual_pointer_v1.cpp:259: Failure
[...]

PR #272, opened 2023-02-02
Update wayland xml

Saviq:


Saviq:
#268 needs fixing first

tests/wlr_virtual_pointer_v1.cpp
```diff
@@ -235,10 +235,35 @@ TEST_F(VirtualPointerV1Test, when_virtual_pointer_scrolls_client_sees_axis)
...
 }
 
-TEST_F(VirtualPointerV1Test, when_virtual_pointer_scrolls_with_steps_client_sees_axis_descrete)
+TEST_F(VirtualPointerV1Test, when_virtual_pointer_scrolls_with_steps_before_wl_pointer_v8_client_sees_only_axis_discrete)
 {
+    // Discrete scrolling is used before wl_pointer v8
+    if (receive_client.bind_if_supported(wl_seat_interface, wlcs::AtLeastVersion(8)))
+    {
+        GTEST_SKIP();
+    }
```
AlanGriffiths, line 244: I don't think `bind_if_supported()` can ever return null?

graysonguarino: It can't return null, but it will skip a test if it fails to bind. What I'm saying here is, "If you can successfully bind, then skip."

AlanGriffiths: Probably needs `wlcs::LessThanVersion(8)` (and some better support in `bind_if_supported()` for that scenario.)

graysonguarino: I now have a working version of `wlcs::LessThanVersion` working, but now that I think about it, I don't think this is exactly what we want in this case (though it may serve use in other cases.)

For this test, the only thing I want to know is the maximum version of an interface that a compositor supports. Something like `auto supports(wl_interface, VersionSpecifier) -> bool` sounds more in line of what we want. I specifically want to know that a compositor updated to wl_pointer v8 does not expect an `axis_discrete()` call.

What's most important, unless I'm thinking of this wrong, is the ability to test that WLCS cannot bind to a particular interface without the behavior of the test being automatically skipped.

RAOF: I'm not sure why we'd need to test that WLCS *can't* bind to a specific version?

If WLCS binds to v8 of the interface, it should get v8 behaviour. If we pass `LessThanVersion{8}` it should bind v7 (or earlier) and get v7 behaviour.

src/in_process_server.cpp
```diff
@@ -1401,6 +1415,7 @@ class wlcs::Client::Impl
         [](auto...){},  // axis_source
         [](auto...){},  // axis_stop
         [](auto...){},  // axis_discrete
+        [](auto...){}, // axis_value120
```
wmww, line 1418: Looks like comment isn't lined up with the others

src/version_specifier.cpp
```diff
@@ -86,4 +86,42 @@ auto wlcs::AtLeastVersion::describe() const -> std::string
...
+wlcs::LessThanVersion::LessThanVersion(uint32_t version) noexcept
+    : version{version}
+{
+}
+
+auto wlcs::LessThanVersion::select_version(
+    uint32_t max_available_version,
+    uint32_t max_supported_version) const -> std::optional<uint32_t>
+{
+    auto const lesser_version = version - 1;
```
wmww, line 98: `lesser_version` is not a particularly meaningful variable name. Perhaps `max_allowed_version` or something would be better?

wmww (changes requested):

src/version_specifier.cpp
```diff
@@ -111,9 +111,9 @@ auto wlcs::LessThanVersion::select_version(
     {
         return {};
     }
-    else if (lesser_version <= max_available_version)
+    else if (max_allowed_version <= max_available_version)
     {
-        return {std::min(lesser_version, max_supported_version)};
+        return {std::min(max_allowed_version, max_supported_version)};
```
RAOF, line 116: Likewise, I think this should be returning the maximum supported version less than the specified version. So, if the compositor supports v5, `LessThanVersion{8}.select_version` should return `{5}`. (This is not critical, since always selecting v1 would *technically* match the behaviour of `LessThanVersion` :wink: )

RAOF (changes requested):

PR #275, opened 2023-02-27
XDG stable: use mock methods instead of notification lists

wmww:
Simplifies and standardizes how events are handled in XDG types

### DIFF
--- a/tests/wlr_foreign_toplevel_management_v1.cpp
+++ b/tests/wlr_foreign_toplevel_management_v1.cpp
@@ -467,10 +467,9 @@ TEST_F(ForeignToplevelHandleTest, gets_activated)
 TEST_F(ForeignToplevelHandleTest, can_maximize_foreign)
 {
     wlcs::XdgToplevelStable::State state{0, 0, nullptr};
-    xdg_toplevel.add_configure_notification(
-        [&state](int32_t width, int32_t height, struct wl_array *states)
+    ON_CALL(xdg_toplevel, configure).WillByDefault([&state](auto... args)
         {
-            state = wlcs::XdgToplevelStable::State{width, height, states};
+            state = wlcs::XdgToplevelStable::State{args...};
         });
     xdg_toplevel_unset_maximized(xdg_toplevel);
     surface.attach_visible_buffer(w, h);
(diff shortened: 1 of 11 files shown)


tests/wlr_foreign_toplevel_management_v1.cpp
```diff
@@ -467,10 +467,9 @@ TEST_F(ForeignToplevelHandleTest, gets_activated)
 TEST_F(ForeignToplevelHandleTest, can_maximize_foreign)
 {
     wlcs::XdgToplevelStable::State state{0, 0, nullptr};
-    xdg_toplevel.add_configure_notification(
-        [&state](int32_t width, int32_t height, struct wl_array *states)
+    ON_CALL(xdg_toplevel, configure).WillByDefault([&state](auto... args)
         {
-            state = wlcs::XdgToplevelStable::State{width, height, states};
+            state = wlcs::XdgToplevelStable::State{args...};
```
RAOF, line 472: Nice use of generic lambdas :)

tests/wlr_foreign_toplevel_management_v1.cpp
```diff
@@ -895,21 +885,15 @@ TEST_F(ForeignToplevelHandleTest, can_activate_foreign)
 
 TEST_F(ForeignToplevelHandleTest, can_close_foreign)
 {
-    bool should_close{false};
-    xdg_toplevel.add_close_notification(
-        [&should_close]()
-        {
-            should_close = true;
-        });
+    EXPECT_CALL(xdg_toplevel, close()).Times(0);
```
RAOF, line 888: I'm not *sure* that this does what you want? GMock expectations are additive; how does this `EXPECT_CALL().Times(0)` interact with the previous `EXPECT_CALL().Times(AnyNumber())`?

RAOF (changes requested):
Yeah, I broadly think this is a much better approach.

Please remove the catch-all expectations, and I'll be most pleased to land this :grin:

tests/wlr_foreign_toplevel_management_v1.cpp
wmww: I believe new more specific `EXPECT_CALL`'s supersede old more general ones. If you remove the `EXPECT_CALL(xdg_toplevel, close()).Times(1);` a few lines down the test fails.

PR #276, opened 2023-03-04
Test popup constraint_adjustment

wmww:
This is a longstanding todo, and a prerequisite for testing the repositioning feature of stable v5. It appears Mir does not respect constraint adjustment, and as such all tests except the slide tests fail.

### DIFF
--- a/tests/xdg_popup.cpp
+++ b/tests/xdg_popup.cpp
@@ -197,7 +197,6 @@ public:
         popup_surface = std::nullopt;
     }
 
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
I started looking at this, I will get back to it in the AM (if it is still pending review)

RAOF (approved):
Took me a bit to convince myself that the expected position was correct for the tests, but (a) I believe they are, and (b) I'm not sure how to make it more obvious.

I note that we're adding another implicit assumption that the edge of the wl_output is the edge of the client-space, but we're not going to clean that up now.

bors r+

PR #283, opened 2023-03-09
[RFC] Copy geometry headers from Mir

wmww:
We've discussed doing this in the past, and the recent work I've done on popup positioning logic got especially annoying without some sort of proper geometry types. The complexity of the Mir ones is arguably overkill for wlcs (some might even argue overkill for Mir), but they're familiar to the Mir team and easy to use. I adapted the layer shell tests to show how they can be used in WLCS, but there's still a lot more work and code churn to fully use them.

### DIFF
--- a/include/geometry/dimensions.h
+++ b/include/geometry/dimensions.h
@@ -14,8 +14,8 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
-#ifndef MIR_GEOMETRY_DIMENSIONS_H_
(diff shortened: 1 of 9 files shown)


AlanGriffiths (commented):
Maybe these should just move to a new package. (Unless we consider mir-core to be sufficiently lightweight?)

RAOF (approved):
My memory of Prague is that everyone was varying degrees of OK with this, and we can pull things out later if necessary.

This nicely simplifies a bunch of things.

PR #291, opened 2023-06-09
Test buffer can be deleted after it's attached

wmww:
Tests https://github.com/MirServer/mir/issues/2960


RAOF (approved):
Yup, "server crash" is definitely the wrong behaviour here!

I've added a comment to that effect, but I don't think that requires re-reviewing.

Issue #293, opened 2023-06-30
VirtualPointerV1Test tests assume server synchonises between send and receive clients


AlanGriffiths:
Specifically, the incantation:
```c++
    send_client.roundtrip();
    receive_client.roundtrip();
```
Is used with the expectation that when `send_client.roundtrip()` returns the corresponding events are available to be processed before the `receive_client.roundtrip()` completes. I don't think there's any reason to expect this.

PR #294, opened 2023-06-30
Less flaky synchronization for VirtualPointerV1Test

AlanGriffiths:
Fixes: #293

### DIFF
--- a/tests/wlr_virtual_pointer_v1.cpp
+++ b/tests/wlr_virtual_pointer_v1.cpp
@@ -23,6 +23,7 @@
 
 #include "linux/input.h"
 #include <gmock/gmock.h>
+#include <thread>
 
 using namespace testing;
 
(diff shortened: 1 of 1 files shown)


tests/wlr_virtual_pointer_v1.cpp
```diff
@@ -95,8 +109,7 @@ class VirtualPointerV1Test: public wlcs::StartedInProcessServer
         EXPECT_CALL(listener, frame()).Times(AnyNumber());
         the_server().move_surface_to(surface, 0, 0);
         pointer.move_to(pointer_start_x, pointer_start_y);
-        send_client.roundtrip();
-        receive_client.roundtrip();
+        wait_for_quiescence();
```
RAOF, line 112: This isn't necessary; we really are just waiting for roundtrips here (there's no virtual pointer usage)

RAOF (changes requested):
So, I think we should just fix this properly. That would be by:
1. Keeping the `send_client.roundtrip()`s in - they're there to ensure that the send-client's queue is flushed to the server (they could be replaced by `wl_client_flush()` I believe)
2. Instead of `receive_client.roundtrip()` actually wait for the events we expect with `receive_client.dispatch_until()` This would be a simple matter of attaching a functor to the correct `EXPECT_CALL` that sets a flag we can wait for.

tests/wlr_virtual_pointer_v1.cpp
```diff
@@ -275,6 +292,7 @@ TEST_F(VirtualPointerV1Test, if_frame_is_not_sent_client_sees_no_events)
     zwlr_virtual_pointer_v1_axis_source(handle, WL_POINTER_AXIS_SOURCE_WHEEL);
     // Should produce no events because there has been no frame
```
RAOF, line 293: This test should actually *test* that no events are produced. As it is, I'm pretty sure this test would pass (but might log some warnings) even if events were generated?

This is pre-existing, though.

RAOF (approved):
Yup! LGTM.

Issue #292, opened 2023-06-20
Missing frame callback in frame_timestamp_increases


bwidawsk:
I've been looking at improving the test coverage in Smithay and just trying to fix failures one by one. For frame_timestamp_increases it appears that only one `frame` request is issued, yet the test expects >= 2 (and why it's not == 2 is kind of sad as well). The proper fix should be to emit two `frame` requests and have a wait for the first to complete before emitting the second.

AlanGriffiths:
This sounds right. From the documentation of `wl_surface_frame()`:

```plain
 * The frame request will take effect on the next wl_surface.commit.
 * The notification will only be posted for one frame unless
 * requested again.
```

> and why it's not == 2 is kind of sad as well

Testing for exactly two frames is a valid thing to test, but it could be misleading if `frame_timestamp_increases` fails for this reason.

PR #295, opened 2023-07-03
Fix various frame-event misunderstandings

RAOF:
Some of our tests have been incorrect because of misunderstandings of the behaviour of various Wayland concepts (damage only applies to pending buffers, a frame callback gets fired at most once, and so on).

Fix these, and while we're at it add an explicit test for the Firefox-endlessly-spins-without-rendering-anything scenario we were kinda accidentally testing (because we were repeatedly sending frame requests without actually submitting anything).

Closes: #292

### DIFF
--- a/tests/frame_submission.cpp
+++ b/tests/frame_submission.cpp
@@ -50,7 +50,8 @@ void FrameSubmission::submit_frame(bool& consumed_flag)
 
     consumed_flag = false;
     wl_callback_add_listener(wl_surface_frame(surface), &frame_listener, &consumed_flag);
-    wl_surface_damage(surface, 0, 0, 200, 200);
+    auto buffer = std::make_shared<wlcs::ShmBuffer>(client, 200, 200);
+    wl_surface_attach(surface, *buffer, 0, 0);
     wl_surface_commit(surface);
 }
 
--- a/tests/frame_submission.cpp
+++ b/tests/frame_submission.cpp
@@ -107,7 +107,7 @@ TEST_F(FrameSubmission, test_buffer_can_be_deleted_after_attached)
  * ends up looping endlessly. If the compositor *doesn't* respond to the frame
  * request, Firefox decides not to draw anything. 🤷‍♀️
  */
-TEST_F(FrameSubmission, test_frame_callback_storm)
+TEST_F(FrameSubmission, when_client_endlessly_requests_frame_then_callbacks_are_throttled)
 {
     using namespace testing;
     using namespace std::chrono_literals;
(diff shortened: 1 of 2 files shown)


tests/frame_submission.cpp
```diff
@@ -98,3 +99,48 @@ TEST_F(FrameSubmission, test_buffer_can_be_deleted_after_attached)
...
 }
+
+/* Firefox has a lovely habit of sending an endless stream of
+ * wl_surface.frame() requests (maybe it's trying to estimate vsync?!)
+ *
+ * If a compositor responds immediately to the frame on commit, Firefox
+ * ends up looping endlessly. If the compositor *doesn't* respond to the frame
+ * request, Firefox decides not to draw anything. 🤷‍♀️
+ */
+TEST_F(FrameSubmission, test_frame_callback_storm)
```
AlanGriffiths, line 110: I don't think this test name tells us much:

    [  FAILED  ] VirtualPointerV1Test.test_frame_callback_storm

```suggestion
TEST_F(FrameSubmission, when_client_endlessly_requests_frame_then_callbacks_are_throttled)
```

tests/test_surface_events.cpp
```diff
@@ -534,11 +534,18 @@ TEST_F(ClientSurfaceEventsTest, frame_timestamp_increases)
     usleep(10000);
 
     wl_surface_attach(surface, buffers[2], 0, 0);
+    surface.add_frame_callback(
+        [&](int time)
+        {
+            EXPECT_THAT(time, Gt(prev_frame_time));
+            prev_frame_time = time;
+            frame_callback_count++;
+        });
```
AlanGriffiths, line 543: Do we need a second copy of the same lambda?

tests/test_surface_events.cpp
```diff
@@ -534,11 +534,18 @@ TEST_F(ClientSurfaceEventsTest, frame_timestamp_increases)
...
     wl_surface_commit(surface);
 
     client.dispatch_until([&frame_callback_count]()
         {
-            return frame_callback_count >= 2;
+            return frame_callback_count == 2;
```
AlanGriffiths, line 548: While we should have a check[*] that `frame_callback_count` is 2, this risks the test hanging if it is more.

[*] Checking this in a test called "frame_timestamp_increases" is wrong: it belongs in "when_two_frames_are_requested_the_two_callbacks_are_seen"

AlanGriffiths (changes requested):
Makes sens, but some nitpicking

tests/test_surface_events.cpp
RAOF: No

tests/test_surface_events.cpp
RAOF: I'm pretty sure it's impossible for `frame_callback_count` to exceed 2. We only call `add_frame_callback` twice, and each callback is one-shot. I'm not *against* making it `>=2`, but no compositor behaviour can cause it to exceed 2.

AlanGriffiths: :shrug:

AlanGriffiths (approved):

PR #298, opened 2023-07-11
Fix frame_timestamp_increases *again*

RAOF:
It turns out that this test was (incorrectly!) marked at XFAIL in Mir's CI. Incorrectly, because Mir actually *does* implement the necessary behaviour.

So Mir CI started to fail, because the XFAIL test was passing. But only sometimes, because it turns out the test was flaky - depending on whether or not `libwayland` felt like actually writing the requests into the display socket or not.

### DIFF
--- a/tests/test_surface_events.cpp
+++ b/tests/test_surface_events.cpp
@@ -530,6 +530,11 @@ TEST_F(ClientSurfaceEventsTest, frame_timestamp_increases)
     surface.add_frame_callback(check_time_and_increment_count);
     wl_surface_commit(surface);
 
+    /* We don't need to wait for the server, but we *do* need
+     * the server to see this commit
+     */
+    client.flush();
+
     /**
      * We need to sleep for multiple miliseconds to make sure the timestamp
      * really does go up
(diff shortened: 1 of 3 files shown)


tests/test_surface_events.cpp
```diff
@@ -530,11 +532,21 @@ TEST_F(ClientSurfaceEventsTest, frame_timestamp_increases)
...
+     * When run against a *real* compositor we should not need any
+     * delay here - when running on a real display, we would expect
+     * the second commit to wait for the next refresh cycle.
+     *
+     * But we're probably not running on a real display, so make
+     * things easier for the integration by waiting a simulated
+     * refresh cycle (at 60Hz) before submitting the next buffer.
      */
-    usleep(10000);
+    std::this_thread::sleep_for(std::chrono::milliseconds{17});
```
mattkae, line 549: **Suggestion**
Do we typically move magic numbers to constants? I am fine either way

mattkae (approved):
One minor suggestion, but makes sense to me!

tests/test_surface_events.cpp
RAOF: How about this?

PR #302, opened 2023-08-16
(#2777) Tests for input-method-v1

mattkae:
# input-method-v1
- **Implementation**: https://github.com/MirServer/mir/pull/3000

## What's new?
- Wrote a bunch of tests for input-method-v1
- I purposefully left out some tests as they are unimplemented in Mir for now, but these will cover some good initial ground

### DIFF
--- a/tests/text_input_v2_with_input_method_v1.cpp
+++ b/tests/text_input_v2_with_input_method_v1.cpp
@@ -172,3 +172,20 @@ TEST_F(TextInputV2WithInputMethodV1Test, input_method_can_send_keysym)
     input_client.roundtrip();
     app_client.roundtrip();
 }
+
+TEST_F(TextInputV2WithInputMethodV1Test, input_method_can_set_preedit_string)
+{
+    auto const preedit_text = "some text";
+    auto const preedit_commit = "some fallback text";
+
+    enable_text_input();
+
+    EXPECT_CALL(text_input, preedit_string(preedit_text, preedit_commit));
+    zwp_input_method_context_v1_preedit_string(
+        *input_method_context,
+        input_method_context->serial,
+        preedit_text,
+        preedit_commit);
+    input_client.roundtrip();
+    app_client.roundtrip();
+}
--- a/tests/text_input_v2_with_input_method_v1.cpp
+++ b/tests/text_input_v2_with_input_method_v1.cpp
@@ -15,9 +15,21 @@
  */
 
 #include "in_process_server.h"
-#include <gmock/gmock.h>
 #include "mock_text_input_v2.h"
 #include "mock_input_method_v1.h"
+#include "version_specifier.h"
+#include "xdg_shell_stable.h"
+#include "method_event_impl.h"
+
+#include <gmock/gmock.h>
+
+// NOTE: In this file, the ordering of app_client.roundtrip() and input_client.roundtrip()
+// is important. If we want the text input to respond to an event triggered by the input method,
+// we should do:
+//    input_client.roundtrip();
+//    app_client.roundtrip();
+// Inversely, we would swap the order if we want the input method to respond to an event triggered
+// by the text input.
 
 using namespace testing;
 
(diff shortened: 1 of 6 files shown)


tests/text_input_v2_with_input_method_v1.cpp
```diff
@@ -0,0 +1,285 @@
...
+    static zwp_input_method_v1_listener constexpr listener {
+        wlcs::method_event_impl<&TextInputV2WithInputMethodV1Test::activate>,
+        wlcs::method_event_impl<&TextInputV2WithInputMethodV1Test::deactivate>
+    };
+
+    void enable_text_input()
+    {
+        create_focused_surface();
+        zwp_text_input_v2_enable(text_input, app_surface->wl_surface());
+        zwp_text_input_v2_update_state(text_input, 0, 0);
```
RAOF, line 74: It would be nice to use the symbolic `update_state` reason here.

Also, I *think* this is wrong? As far as I can tell:
* You don't need to send an `update_state` in order for `enable` to be processed - the protocol is unclear, but it doesn't *say* `enable` is double-buffered state committed by `update_state`, and it's not important that it be associated with a particular batch of changes
* The serial here should be from an `enter` (or `input_method_changed`) event. We should have got one of those as a part of the surface being focused.

tests/text_input_v2_with_input_method_v1.cpp
```diff
@@ -0,0 +1,285 @@
...
+    EXPECT_THAT(entered, Eq(app_surface.value().wl_surface()));
+}
+
+TEST_F(TextInputV2WithInputMethodV1Test, text_input_activates_context_on_enable)
+{
+    create_focused_surface();
+    zwp_text_input_v2_enable(text_input, app_surface->wl_surface());
+    zwp_text_input_v2_update_state(text_input, 0, 0);
+    app_client.roundtrip();
+    input_client.roundtrip();
```
RAOF, line 104: These cross-client synchronisation requirements are hard; it's possible this is racy (I don't think it is in Mir, but I don't believe the protocol guarantees it isn't).

Probably you want to do `app_client.roundtrip()` and then `input_client.dispatch_until(...)` for the input client.

tests/text_input_v2_with_input_method_v1.cpp
```diff
@@ -0,0 +1,285 @@
...
+{
+    create_focused_surface();
+
+    EXPECT_CALL(*this, deactivate(_));
+    zwp_text_input_v2_enable(text_input, app_surface->wl_surface());
+    zwp_text_input_v2_update_state(text_input, 0, 0);
+    app_client.roundtrip();
+    input_client.roundtrip();
+    zwp_text_input_v2_disable(text_input, app_surface->wl_surface());
+    zwp_text_input_v2_update_state(text_input, 1, 0);
```
RAOF, line 118: Again, I'm reasonably confident that `enable`/`disable` does not require committing with `update_state`.

tests/text_input_v2_with_input_method_v1.cpp
```diff
@@ -0,0 +1,285 @@
...
+}
+
+TEST_F(TextInputV2WithInputMethodV1Test, input_method_can_set_modifiers_map)
+{
+    // Note: This example data was taken from the maliit-keyboard example
+    auto const text = "hello";
+    const size_t data_length = 33;
+    wl_array map;
+    wl_array_init(&map);
+    const char** data = static_cast<const char**>(wl_array_add(&map, sizeof(char) * data_length));
```
RAOF, line 257: Style nit: You can probably declare this as `auto data`, as you're explicitly casting on the rhs?

RAOF (changes requested):
This looks pretty good all up.

The two pervasive issues are cross-client synchronisation (which should *probably* use `dispatch_until` on the receiving end), and serial handling for `update_state`. The latter is important because a compositor could reasonably discard any `update_state` requests with an incorrect serial - after all, they're attempting to update no-longer-valid state.

tests/text_input_v2_with_input_method_v1.cpp
mattkae: I _think_ I did this correctly now :thinking: Let me know!

RAOF (approved):
LGTM now!

PR #313, opened 2023-11-10
wlcs::Client: Handle incomplete logical pointer/touch events better.

RAOF:
A sequence of `wl_pointer`/`wl_touch` events are combined into a logical event by the terminating `frame` event. When tests ask for the current touch/pointer position, or window under touch/pointer, always return the current state, even if there is an incomplete logical event (ie: some events have been received, but not a terminating `frame` event).

There's no guarantee that a compositor will emit a complete logical event in a single flush of the protocol socket, so there's no guarantee that we won't run test code after receiving (for example) a `wl_pointer.enter` event but before processing a `wl_pointer.frame`.

So it is incorrect to fail the test in such cases

### DIFF
--- a/src/in_process_server.cpp
+++ b/src/in_process_server.cpp
@@ -878,8 +878,6 @@ public:
 
     wl_surface* window_under_cursor() const
     {
-        if (pointer_events_pending())
-            BOOST_THROW_EXCEPTION(std::runtime_error("Pointer events pending"));
         if (current_pointer_location)
         {
             return current_pointer_location->surface;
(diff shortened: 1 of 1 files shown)


RAOF:
See also: [this exchange in #Mir](https://matrix.to/#/!ivAdrrLoWkBYEnSIae:matrix.org/$fASWmRVc6_XtMHcTjZICKg5gZL8obGkdB7Cg-EW6Vu0?via=matrix.org&via=libera.chat&via=cooperteam.net)

RAOF:
See also: [this exchange in #Mir](https://matrix.to/#/!ivAdrrLoWkBYEnSIae:matrix.org/$fASWmRVc6_XtMHcTjZICKg5gZL8obGkdB7Cg-EW6Vu0?via=matrix.org&via=libera.chat&via=cooperteam.net)

AlanGriffiths (approved):
I'm convinced

PR #315, opened 2023-11-14
debian: drop gcc-12 workaround

Saviq:
The remaining issue is #314, with which this doesn't help anyway.


Saviq:
This failed like #314 regardless of the workaround:

https://launchpad.net/~mir-team/+archive/ubuntu/test/

Issue #318, opened 2023-11-29
ConfigurationWindow constructor is racy


bwidawsk:
`ConfigurationWindow` waits for configure to be sent by the display server. However, the check/wait is only initialized after the client roundtrip, which means the configure may have already been sent, and this, the check logic will never trigger.

Proposed fix:
```
diff --git a/tests/xdg_toplevel_stable.cpp b/tests/xdg_toplevel_stable.cpp
index 676fc6b482a7..6535cbc75ff9 100644
--- a/tests/xdg_toplevel_stable.cpp
+++ b/tests/xdg_toplevel_stable.cpp
@@ -49,6 +49,7 @@ public:
           xdg_shell_surface{client, surface},
           toplevel{xdg_shell_surface}
     {
+       int prev = surface_configure_count;
         ON_CALL(xdg_shell_surface, configure).WillByDefault([&](auto serial)
             {
                 xdg_surface_ack_configure(xdg_shell_surface, serial);
@@ -64,16 +65,21 @@ public:
         client.roundtrip();
         surface.attach_buffer(window_width, window_height);
         wl_surface_commit(surface);
-        dispatch_until_configure();
+        __dispatch_until_configure(prev);
+    }
+
+    void __dispatch_until_configure(int prev)
+    {
+        client.dispatch_until(
+            [prev_count = prev, &current_count = surface_configure_count]()
+            {
+                return current_count > prev_count;
+            });
     }
 
     void dispatch_until_configure()
     {
-        client.dispatch_until(
-            [prev_count = surface_configure_count, &current_count = surface_configure_count]()
-            {
-                return current_count > prev_count;
-            });
+       __dispatch_until_configure(surface_configure_count);
     }
 
 ```

RAOF:
I've slightly reworked that fix for aesthetic preference; does the linked PR work for you?

RAOF:
Hm, actually, I think the problem you're hitting is that we expect a *second* configure after the buffer is committed that sets the `activated` state. I guess we assume that prior to being mapped the surface is *not* in the `activated` state.

RAOF:
I've updated that PR *again*; Mir passes this, and it matches the behaviour I see under GNOME Shell, but it's maybe not exactly behaviour *mandated* by the protocol.

PR #321, opened 2023-12-03
XdgToplevelStable: Accept more protocol-compliant behaviour.

RAOF:
We assumed that the compositor would send a second `xdg_surface.configure` after committing a buffer, as now that the surface is mapped we expect it to be in the `activated` state.

This second configure isn't guaranteed by the protocol, and there's no reason a compositor couldn't set the `activated` state in the *first* `configure`, prior to the window being mapped, on the basis that it'll almost certainly be made active once it *has* been mapped¹.

Instead of waiting for a second configure, wait for the surface to be in `activated` state. That catches both, equally protocol compliant, compositor behaviours.

Fixes: #318 (again!)

¹: This is possibly desirable behaviour - this way the client can immediately draw in the activated state, rather than rendering for an inactive window and then immediately needing to re-render for an active window.

### DIFF
--- a/tests/xdg_toplevel_stable.cpp
+++ b/tests/xdg_toplevel_stable.cpp
@@ -75,12 +75,17 @@ public:
 
         surface.attach_buffer(window_width, window_height);
         wl_surface_commit(surface);
+        client.flush();
 
         /* Now that we've committed a buffer (and hence should be mapped) we expect
-         * to be reconfigured with new state - particularly, we expect to be
-         * in “activated” state after this.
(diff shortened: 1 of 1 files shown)


bwidawsk:
I've reworked my compositor to send a second configure with activated state after the buffer is attached. While this is not required per the spec, it is the only feasible way for me to set activated state (which apparently WLCS relies on). As such, this PR is no longer needed.

AlanGriffiths (approved):

Issue #312, opened 2023-11-09
XDG configure event expected without commit


bwidawsk:
https://wayland.app/protocols/xdg-shell#xdg_surface

> For an xdg_surface to be mapped by the compositor, the following conditions must be met: (1) the client has assigned an xdg_surface-based role to the surface (2) the client has set and committed the xdg_surface state and the role-dependent state to the surface (3) the client has committed a buffer to the surface

Tests such as XdgSurfaceStableTest.gets_configure_event do not actually commit while expecting configure.

Something like the following should make correct compositors pass:
``` diff
diff --git a/tests/xdg_surface_stable.cpp b/tests/xdg_surface_stable.cpp
index c0a543777a1a..2b50389724cb 100644
--- a/tests/xdg_surface_stable.cpp
+++ b/tests/xdg_surface_stable.cpp
@@ -58,6 +58,7 @@ TEST_F(XdgSurfaceStableTest, gets_configure_event)
         });

     wlcs::XdgToplevelStable toplevel{xdg_surface};
+    wl_surface_commit(surface);
     surface.attach_buffer(600, 400);

     client.roundtrip();
```

AlanGriffiths:
I don't disagree with your conclusion, but I think the correct reference is:

> After creating a role-specific object and setting it up, the client must perform an initial commit without any buffer attached. The compositor will reply with initial wl_surface state such as wl_surface.preferred_buffer_scale followed by an xdg_surface.configure event.

_(ibid)_

PR #322, opened 2023-12-04
XdgSurfaceStable: Fix configure event logic

RAOF:
We have a test, further down the file, testing that attaching a buffer to an `xdg_surface` before the first commit throws a protocol error.

Fix `gets_configure_event` to not attach a buffer to an `xdg_surface` before the first commit 🤦‍♀️.

Fixes: #312


AlanGriffiths (approved):
LGTM

PR #324, opened 2023-12-08
InProcessServer: Fix xdg_shell window construction.

RAOF:
Protocol requires that a client `commit` the surface after setting up XDG roles and `ack_configure` the resulting `xdg_surface.configure` event *before* attaching a buffer to the surface.

We don't notice in Mir because we XFAIL the relevant tests 🤦‍♀️.


AlanGriffiths (approved):
LGTM

PR #333, opened 2024-03-13
(#332) LayerSurfaceLayoutTest includes a test for a -1 exclusion zone

### LINKED ISSUE #332: `LayerSurfaceLayoutTest` should have a test for a `-1` exclusion zone
## Request
Per the spec (https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_surface_v1:request:set_exclusive_zone), doing `zwlr_layer_surface_v1::set_exclusive_zone(-1)` should result in a surface which disregards all other exclusion zones. For example, `swaybg` uses this to ensure that the background takes up the full size of the display area.

mattkae:
fixes #332 

Implemented a test for `zwlr_layer_surface_v1_set_exclusive_zone(-1)`, which means that the surface should ignore all exclusion zones (e.g. a background)

### DIFF
--- a/tests/wlr_layer_shell_v1.cpp
+++ b/tests/wlr_layer_shell_v1.cpp
@@ -949,7 +949,6 @@ TEST_P(LayerSurfaceLayoutTest, surfaces_with_exclusive_zone_set_to_negative_one_
     auto const layout = GetParam();
     auto const request_size = layout.request_size();
     auto const initial_rect = layout.placement_rect(output_rect());
-    auto const exclusive = 12;
 
     zwlr_layer_surface_v1_set_anchor(layer_surface, layout);
     zwlr_layer_surface_v1_set_exclusive_zone(layer_surface, -1);
@@ -958,9 +957,17 @@ TEST_P(LayerSurfaceLayoutTest, surfaces_with_exclusive_zone_set_to_negative_one_
     commit_and_wait_for_configure();
     surface.attach_visible_buffer(initial_rect.size.width.as_int(), initial_rect.size.height.as_int());
 
-    // Create layer surfaces with exclusive zones on the top and left of the output that would
-    // theoretically push our surface out of the way if it did NOT have an exclusive zone of -1.
+    // First, let's check that the surface is in the position that we expect and
+    // that the compositor has suggested the size that we expect.
+    client.roundtrip();
 
+    EXPECT_THAT(layer_surface.last_size().width, Eq(initial_rect.size.width));
+    EXPECT_THAT(layer_surface.last_size().height, Eq(initial_rect.size.height));
+    expect_surface_is_at_position(initial_rect.top_left);
+
+    // Next, create layer surfaces with exclusive zones on the top and left of the output that would
+    // theoretically push our surface out of the way if it did NOT have an exclusive zone of -1.
+    auto const exclusive = 12;
     wlcs::Surface top_surface{client};
     wlcs::LayerSurfaceV1 top_layer_surface{client, top_surface};
     zwlr_layer_surface_v1_set_anchor(top_layer_surface, ZWLR_LAYER_SURFACE_V1_ANCHOR_TOP);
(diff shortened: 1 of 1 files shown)


tests/wlr_layer_shell_v1.cpp
```diff
@@ -944,6 +944,61 @@ TEST_P(LayerSurfaceLayoutTest, maximized_xdg_toplevel_is_shrunk_for_exclusive_zo
...
+    {
+        EXPECT_THAT(layer_surface.last_size().width, Eq(expected_config_size.width));
+    }
+    if (expected_config_size.height.as_int())
+    {
+        EXPECT_THAT(layer_surface.last_size().height, Eq(expected_config_size.height));
+    }
+
+    auto const expected_placement = layout.placement_rect(non_exlusive_zone);
+    surface.attach_visible_buffer(expected_placement.size.width.as_int(), expected_placement.size.height.as_int());
```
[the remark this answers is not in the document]
RAOF: Additionally, you're never `commit`ting this buffer, so it's not going to do anything?

tests/wlr_layer_shell_v1.cpp
```diff
@@ -944,6 +944,61 @@ TEST_P(LayerSurfaceLayoutTest, maximized_xdg_toplevel_is_shrunk_for_exclusive_zo
...
+    if (expected_config_size.width.as_int())
+    {
+        EXPECT_THAT(layer_surface.last_size().width, Eq(expected_config_size.width));
+    }
+    if (expected_config_size.height.as_int())
+    {
+        EXPECT_THAT(layer_surface.last_size().height, Eq(expected_config_size.height));
+    }
+
+    auto const expected_placement = layout.placement_rect(non_exlusive_zone);
```
RAOF, line 997: Why are we naming this differently? `expected_placement == initial_rect`

RAOF (changes requested):
I'm a bit confused at what this test is doing, and I think it's because the test isn't testing what you think it is.

I think it would be more obvious what it's doing if either:
* You set up the exclusive surfaces first, then set up the `exclusive=-1` surface and checked its position, or
* You set up the `exclusive=-1` surface, checked its position, *then* set up the exclusive surfaces and asserted that no reconfiguration occured.

(It's possible that this is more usefully *two* tests)

mattkae:
> I'm a bit confused at what this test is doing, and I think it's because the test isn't testing what you think it is.

I am attempting to test that ignoring exclusive zones while you are a fullscreened surface gives you the whole output, regardless of there being attached surfaces with exclusive zones defined.

> I think it would be more obvious what it's doing if either:
> 
>     * You set up the exclusive surfaces first, then set up the `exclusive=-1` surface and checked its position, or
> 
>     * You set up the `exclusive=-1` surface, checked its position, _then_ set up the exclusive surfaces and asserted that no reconfiguration occured.
> 
> 
> (It's possible that this is more usefully _two_ tests)

These two tests sound like a better idea tho

tests/wlr_layer_shell_v1.cpp
```diff
@@ -944,6 +944,65 @@ TEST_P(LayerSurfaceLayoutTest, maximized_xdg_toplevel_is_shrunk_for_exclusive_zo
...
+    commit_and_wait_for_configure();
+    surface.attach_visible_buffer(initial_rect.size.width.as_int(), initial_rect.size.height.as_int());
+
+    // First, let's check that the surface has the size and position that we would expect
+    client.roundtrip();
+
+    if (request_size.width.as_int())
+        EXPECT_THAT(layer_surface.last_size().width, Eq(request_size.width));
+    if (request_size.height.as_int())
+        EXPECT_THAT(layer_surface.last_size().height, Eq(request_size.height));
```
RAOF, line 966: I'm pretty sure this isn't testing what you expect it to be testing? Particularly: we've attached a buffer of size `initial_rect.size`. That means that the surface has size `initial_rect.size`; the client buffer size is authoritative.

What this is testing is that the WM has *suggested* (in the `configure` event) that we make the surface size `request_size.size`.

RAOF (changes requested):
Let's do some synchronous review of this next week. :wink:

RAOF (approved):
Yeah, this is clearer now. Sweet!

PR #339, opened 2024-06-21
Add tests for xdg-decoration-unstable-v1.

tarek-y-ismail:


tests/xdg_decoration.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+namespace
+{
+auto const any_width = 300;
+auto const any_height = 300;
+
+struct ToplevelDecoration : StartedInProcessServer
+{
+    Client a_client{the_server()};
+    Surface a_surface{a_client.create_xdg_shell_stable_surface(any_width, any_height)};
```
AlanGriffiths, line 35: This incantation hides the `xdg_get_toplevel`. If you need to access that, there is `XdgSurfaceStable`. See, for example, `ForeignToplevelHandleTest`

src/in_process_server.cpp
```diff
@@ -1938,6 +1938,12 @@ class wlcs::Surface::Impl
         owner_.dispatch_until([surface_rendered]() { return *surface_rendered; });
     }
 
+    void attach_toplevel(xdg_toplevel* toplevel) {
+	    this->toplevel = toplevel;
+    }
```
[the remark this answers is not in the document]
tarek-y-ismail: Indeed I wasn't very comfortable doing this. Your other comment showed me a better way to be able to access the data I needed. Will try grepping around more in the future. And for the formatting, I should've caught this before pushing, apologies. I'll try to focus on formatting more in the future.

tests/xdg_decoration_v1.cpp
AlanGriffiths, line 1: We are missing tests that `set_mode()` and `unset_mode()` requests elicit a `configure()` event.

RAOF (commented):
A couple of comments.

tests/xdg_decoration_v1.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+TEST_F(XdgDecorationV1Test, already_constructed)
+{
+    {
+        XdgToplevelStable xdg_toplevel{xdg_surface};
+        ZxdgToplevelDecorationV1 decoration{manager, xdg_toplevel};
+        ZxdgToplevelDecorationV1 duplicate_decoration{manager, xdg_toplevel};
+    }
+
+    EXPECT_THROW(a_client.roundtrip(), wlcs::ProtocolError);
```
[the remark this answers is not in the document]
tarek-y-ismail: Done! In fact, it helped me catch a "bug" where the interface didn't match what I expected (was `zxdg_decoration_manager_v1_interface` while I expected `zxdg_toplevel_decoration_v1_interface`, but given where the errors are reported, the former is more fitting) 

Thanks a lot :)

tests/xdg_decoration_v1.cpp
```diff
@@ -0,0 +1,83 @@
...
+    ZxdgToplevelDecorationV1 decoration{manager, xdg_toplevel};
+
+    zxdg_toplevel_decoration_v1_set_mode(decoration, ZXDG_TOPLEVEL_DECORATION_V1_MODE_CLIENT_SIDE);
+    zxdg_toplevel_decoration_v1_set_mode(decoration, ZXDG_TOPLEVEL_DECORATION_V1_MODE_SERVER_SIDE);
+    zxdg_toplevel_decoration_v1_unset_mode(decoration);
+
+    EXPECT_CALL(decoration, configure(_)).Times(3);
+
+    a_client.roundtrip();
+}
```
AlanGriffiths, line 83: Three separate tests would be both clearer (if the test fails what response(s) was missing) and more likely to pick up problems. (E.g. `set_mode()` doing nothing and `unset_mode()`  getting three responses)

tests/xdg_decoration_v1.cpp
```diff
@@ -0,0 +1,83 @@
...
+TEST_F(XdgDecorationV1Test, set_and_unset_mode_result_in_a_config_event)
+{
+    XdgToplevelStable xdg_toplevel{xdg_surface};
+    ZxdgToplevelDecorationV1 decoration{manager, xdg_toplevel};
+
+    zxdg_toplevel_decoration_v1_set_mode(decoration, ZXDG_TOPLEVEL_DECORATION_V1_MODE_CLIENT_SIDE);
+    zxdg_toplevel_decoration_v1_set_mode(decoration, ZXDG_TOPLEVEL_DECORATION_V1_MODE_SERVER_SIDE);
+    zxdg_toplevel_decoration_v1_unset_mode(decoration);
+
+    EXPECT_CALL(decoration, configure(_)).Times(3);
```
AlanGriffiths, line 80: While the server is required to respond to each request, I the protocol allows `configure()` "at any time":
```suggestion
    EXPECT_CALL(decoration, configure(_)).Times(AtLeast(3));
```

AlanGriffiths (changes requested):
Some small tweaking required

tests/xdg_decoration_v1.cpp
tarek-y-ismail: I'm not sure I understand, the main issue you're trying to convey is the response being sent at a time where gtest doesn't track it (too early/late), correct? But what I understand is that `AtLeast` would allow for more than 3 `configure` calls. So I'm not sure how that helps us deal with the issue

AlanGriffiths: The issue is that a server that sends more `configure()` events than needed should still pass the test

tests/xdg_decoration_v1.cpp
```diff
@@ -68,16 +68,38 @@ TEST_F(XdgDecorationV1Test, destroying_toplevel_before_decoration_throws_orphane
...
 
-TEST_F(XdgDecorationV1Test, set_and_unset_mode_result_in_a_config_event)
+TEST_F(XdgDecorationV1Test, set_mode_client_results_in_a_configure_event)
 {
     XdgToplevelStable xdg_toplevel{xdg_surface};
     ZxdgToplevelDecorationV1 decoration{manager, xdg_toplevel};
 
     zxdg_toplevel_decoration_v1_set_mode(decoration, ZXDG_TOPLEVEL_DECORATION_V1_MODE_CLIENT_SIDE);
+
+    EXPECT_CALL(decoration, configure(_)).Times(1);
```
AlanGriffiths, line 78:
```suggestion
    EXPECT_CALL(decoration, configure(_)).Times(AtLeast(1));
```

AlanGriffiths (approved):
Okay

PR #341, opened 2024-07-24
debian: drop custom LTO handling

Saviq:
Leave it to the distro instead.


RAOF (approved):
Same justification as Mir, yeah.

PR #342, opened 2024-08-05
Add tests for fractional-scale-v1

tarek-y-ismail:
Currently tests:
- The happy path
- Duplicate fractional scaling components


include/fractional_scale_v1.h
RAOF, line 1: Hm. I think this is unnecessary infrastructure. We don't use any of it, other than to construct the Wayland objects, and I think there it serves to make things less clear than just calling the `wp_fractional_scale_manager_v1_get_fractional_scale(...)` function directly.

`wlcs::WpFractionalScaleV1` might be interesting later¹, if we manage to add enough server hooks so that we can usefully test the events, but for now I think it's better not in here.

¹: Because the interface can receive events it *is* useful to have a wrapper around the Wayland object to set the listener up and delegate to `MOCK_METHOD`s

include/fractional_scale_v1.h
```diff
@@ -0,0 +1,66 @@
...
+    ~WpFractionalScaleV1();
+
+    MOCK_METHOD(void, preferred_scale, (uint32_t scale), ());
+
+    operator wp_fractional_scale_v1*() const;
+    friend void wp_fractional_scale_v1_destroy(WpFractionalScaleManagerV1 const&) = delete;
+
+private:
+    uint32_t const version;
+    wp_fractional_scale_v1* const fractional_scale;
```
RAOF, line 59: Should this be a `WlHandle<wp_fractional_scale_v1>`? The object has a `destroy` request.

RAOF (changes requested):

tests/fractional_scale_v1.cpp
```diff
@@ -0,0 +1,49 @@
...
+    Client a_client{the_server()};
+    WpFractionalScaleManagerV1 manager{a_client};
+    Surface a_surface{a_client};
+};
+
+TEST_F(FractionalScaleV1Test, fractional_scale_creation_does_not_throw)
+{
+    WpFractionalScaleV1 fractional_scale{manager, a_surface};
+
+    EXPECT_NO_THROW(a_client.roundtrip());
```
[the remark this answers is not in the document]
RAOF: I think this is slightly *more* readable without the wrapper class infrastructure?

include/fractional_scale_v1.h
tarek-y-ismail: You're correct. A lot of boilerplate was removed!

include/fractional_scale_v1.h
tarek-y-ismail: Yes, both `wp_fractional_scale_v1` and `wp_fractional_scale_manager_v1` have destroy requests.  I didn't really pay attention to them since the tests aren't affected by them (afaik) but good wayland etiquette never hurts. 

Thanks for pointing this out!

tests/fractional_scale_v1.cpp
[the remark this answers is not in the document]
tarek-y-ismail: I agree. I was planning on removing the wrapper class and seeing how things would look when I had downtime. You pointing this out was even more of a reason to try it out. Thank you

tests/fractional_scale_v1.cpp
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright © 2020 Canonical Ltd.
```
RAOF, line 2: We're not putting dates on our copyright. Also, are you a time traveller? :wink:

tests/fractional_scale_v1.cpp
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+#include "expect_protocol_error.h"
+#include "generated/fractional-scale-v1-client.h"
+#include "in_process_server.h"
+#include "version_specifier.h"
+#include "wl_handle.h"
+
+using namespace wlcs;
+using testing::Eq;
```
RAOF, line 24: Don't really need this anymore?

RAOF (approved):
Yup, looks good. Just a couple of minor nits; feel free to fix them, commit, and then hit the merge button!

tests/fractional_scale_v1.cpp
tarek-y-ismail: Only the include is not needed. the namespace helps a bit with code nose and `testing::Eq` is required for `EXPECT_PROTOCOL_ERROR`.

PR #343, opened 2024-09-05
Add (non-working) test to reproduce QT's resize pingpong bug

tarek-y-ismail:
Right now, the test doesn't actually reproduce the bug. My suspicion is the code called in `WillRepeatedly`. Maybe a couple more eyes will help me pinpoint the issue.


tarek-y-ismail:
Still can't seem to make the bug happen. @canonical/mir any QT quirks that might be contributing to this?

PR #345, opened 2024-10-14
Fix use after free in WpViewporter utility function

tarek-y-ismail:
Extends the lifetime of the variables captured by the lambda to work aound use-after-free ASAN errors.

### DIFF
--- a/include/in_process_server.h
+++ b/include/in_process_server.h
@@ -272,7 +272,6 @@ public:
     void add_pointer_enter_notification(PointerEnterNotifier const& on_enter);
     void add_pointer_leave_notification(PointerLeaveNotifier const& on_leave);
     void add_pointer_motion_notification(PointerMotionNotifier const& on_motion);
-    void clear_pointer_motion_notifications();
     void add_pointer_button_notification(PointerButtonNotifier const& on_button);
 
     void dispatch_until(
--- a/include/in_process_server.h
+++ b/include/in_process_server.h
@@ -272,6 +272,7 @@ public:
     void add_pointer_enter_notification(PointerEnterNotifier const& on_enter);
     void add_pointer_leave_notification(PointerLeaveNotifier const& on_leave);
     void add_pointer_motion_notification(PointerMotionNotifier const& on_motion);
+    void clear_pointer_motion_notifications();
     void add_pointer_button_notification(PointerButtonNotifier const& on_button);
 
     void dispatch_until(
--- a/tests/wp_viewporter.cpp
+++ b/tests/wp_viewporter.cpp
@@ -112,12 +112,12 @@ public:
                 pointer_entered = false;
                 return false;
             });
-        client.add_pointer_motion_notification(
-            [&](auto, auto)
-            {
-                motion_received = true;
-                return false;
-            });
+
+        client.add_pointer_motion_notification([&](auto, auto)
+        {
+            motion_received = true;
+            return false;
+        });
 
         pointer.move_by(1, 1);
         client.dispatch_until([&]() { return !pointer_entered || motion_received; });
@@ -128,6 +128,7 @@ public:
             return false;
         }
 
+        client.clear_pointer_motion_notifications();
         return true;
     }
 };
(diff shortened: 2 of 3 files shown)


AlanGriffiths (changes requested):
> More of a band-aid solution, really. There's no quick way to remove individual std::function listeners so we clear them all out.

I think this is likely to cause problems as clearing _all_ "listeners" is surprising behaviour.

A better approach would be to fix the captures so they do not reference local variables.

There are several design options:

1. Move the variables to class scope
1.1.  They could be members of `WpViewporterTest`, but then their close connection to `surface_has_size()` would be unclear; or,
1.2. Move `surface_has_size()` to a helper class (would need to add `the_server()` to the parameters or constructor.
2. Share ownership of the variables:

```c++
    auto const pointer_entered = std::make_shared<bool>(false);
...
        client.add_pointer_enter_notification(
            [pointer_entered](auto entered_surface, auto, auto)
            {
                *pointer_entered = ...;
                return false;
            });
...

        client.dispatch_until([&]() { return ... !*pointer_entered; });
```

(The variables could also be scoped to the fixture, but then the scope of usage becomes less clear.)

include/in_process_server.h
```diff
@@ -272,6 +272,7 @@ class Client
     void add_pointer_enter_notification(PointerEnterNotifier const& on_enter);
     void add_pointer_leave_notification(PointerLeaveNotifier const& on_leave);
     void add_pointer_motion_notification(PointerMotionNotifier const& on_motion);
+    void clear_pointer_motion_notifications();
```
AlanGriffiths, line 275:
```suggestion
```

tests/wp_viewporter.cpp
```diff
@@ -128,6 +132,7 @@ class WpViewporterTest : public wlcs::InProcessServer
             return false;
         }
 
+        client.clear_pointer_motion_notifications();
```
AlanGriffiths, line 135:
```suggestion
```

AlanGriffiths (changes requested):

tests/wp_viewporter.cpp
```diff
@@ -104,21 +108,21 @@ class WpViewporterTest : public wlcs::InProcessServer
...
             });
 
+        client.add_pointer_motion_notification([motion_received](auto, auto)
+        {
+            *motion_received = true;
+            return false;
+        });
+
         pointer.move_by(1, 1);
         client.dispatch_until([&]() { return !pointer_entered || motion_received; });
```
AlanGriffiths, line 127: The trouble with pointers is that they implicitly convert to bool.
```suggestion
        client.dispatch_until([&]() { return !*pointer_entered || *motion_received; });
```

tarek-y-ismail:
Converted to draft until I get compile errors fixed and test on a CI machine

tests/wp_viewporter.cpp
```diff
@@ -47,20 +48,23 @@ class WpViewporterTest : public wlcs::InProcessServer
 
         auto pointer = the_server().create_pointer();
 
-        bool pointer_entered = false;
-        bool motion_received = false;
+        // Use a shared ptr to extend the lifetime of the variables until the
+        // pointer/motion notifications are done with them. Otherwise, we get a
+        // use-after-free.
+        auto pointer_entered = std::make_shared<bool>(false);
+        auto motion_received = std::make_shared<bool>(false);
```
AlanGriffiths, line 55: These can, and should be `const`

tests/wp_viewporter.cpp
```diff
@@ -104,23 +108,23 @@ class WpViewporterTest : public wlcs::InProcessServer
...
-                motion_received = true;
+                *pointer_entered = false;
                 return false;
             });
 
+        client.add_pointer_motion_notification([motion_received](auto, auto)
+        {
+            *motion_received = true;
+            return false;
+        });
```
AlanGriffiths, line 124: Why reformat this lambda? You didn't see a need to reformat the other three

AlanGriffiths (changes requested):
Nits

tests/wp_viewporter.cpp
tarek-y-ismail: Forgot to format it in the rush to get CI fixed. Should be all good now.

AlanGriffiths (approved):
Good now

PR #350, opened 2025-02-07
FrameSubmission: Expand test for Firefox frame storm.

RAOF:
The existing test checks that we don't generate a frame storm, but does *not* check that we actually send a frame callback, so simply dropping any of these requests on the floor passes.

Add an extra test checking that we actually send these frame callbacks.

### DIFF
--- a/tests/frame_submission.cpp
+++ b/tests/frame_submission.cpp
@@ -16,10 +16,8 @@
  * Authored by: Alan Griffiths <alan@octopull.co.uk>
  */
 
-#include "helpers.h"
 #include "in_process_server.h"
 
-#include <functional>
 #include <gmock/gmock.h>
 
 using namespace testing;
--- a/tests/frame_submission.cpp
+++ b/tests/frame_submission.cpp
@@ -153,7 +153,7 @@ TEST_F(FrameSubmission, when_client_endlessly_requests_frame_then_callbacks_are_
  */
 TEST_F(FrameSubmission, frame_requests_without_buffer_are_called_back_eventually)
 {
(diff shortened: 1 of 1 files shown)


tests/frame_submission.cpp
```diff
@@ -19,6 +19,7 @@
 #include "helpers.h"
 #include "in_process_server.h"
 
+#include <functional>
```
AlanGriffiths, line 22: The new code doesn't seem to need this, so why?

tests/frame_submission.cpp
```diff
@@ -144,3 +145,26 @@ TEST_F(FrameSubmission, when_client_endlessly_requests_frame_then_callbacks_are_
...
+/*
+ * The above test checks that frame callbacks are throttled, but not that
+ * they are sent at all.
+ *
+ * We also need to test that these type of buffer-less frame requests
+ * *do* get a callback, eventually.
+ */
+TEST_F(FrameSubmission, frame_requests_without_buffer_are_called_back_eventually)
+{
+    std::array<bool, 5> called = {0};
```
AlanGriffiths, line 158: Relying on the implicit conversion to `bool` seems ugly
```suggestion
    std::array<bool, 5> called{false};
```

AlanGriffiths (approved):
A couple of optional nits

tests/frame_submission.cpp
RAOF: I think our code is being misleading in both cases; this is aggregate initialisation, right? Which is going to initialised the *first* member with the value we provide, then default-initialise the rest of the array (conveniently, in this case, to `false`)?

PR #352, opened 2025-05-11
Use the latest GPLv2 license text

musicinmybrain:
Replaces the contents of `COPYING.GPL2` with the latest https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.

- FSF is now remote-only, so there is no longer a postal address
- A placeholder name has been changed


robert-ancell (approved):
Thanks!

PR #354, opened 2025-06-19
Add helpers for "how long should a timeout be"

RAOF:
And then use this to make those timeouts *architecture specific*,
since RISCV is so slow that it often hits existing timeouts.


RAOF:
(Partially) Fixes https://github.com/canonical/mir/issues/3996

PR #358, opened 2025-06-24
use_a_short_time_more

AlanGriffiths:
ClientSurfaceEventsTest.frame_timestamp_increases's "wait" logic was wrong.

We cannot expect strictly monotonic timestamps unless we wait for the first frame callback: There is no way for the client to know the state of the server.

(Only if there were no need for a wait could the test be correct)

So, split the test in two: 

1. wait for the first frame callback and assert timestamps increases
2. don't wait and assert timestamps do not decrease


Saviq (approved):
ACK

PR #359, opened 2025-07-16
Add first step for linux_dmabuf tests

robert-ancell:


RAOF (changes requested):
Feedback inline.

include/linux_dmabuf_v1.h
```diff
@@ -0,0 +1,68 @@
...
+    LinuxDmabufFeedbackV1& operator=(LinuxDmabufFeedbackV1 const&) = delete;
+    ~LinuxDmabufFeedbackV1();
+
+    MOCK_METHOD0(done, void());
+    MOCK_METHOD2(format_table, void(int fd, uint32_t size));
+    MOCK_METHOD1(main_device, void(dev_t devnum));
+    MOCK_METHOD0(tranche_done, void());
+    MOCK_METHOD1(tranche_target_device, void(dev_t devnum));
+    MOCK_METHOD1(tranche_formats, void(std::vector<uint32_t> indices));
+    MOCK_METHOD1(tranche_flags, void(uint32_t flags));
```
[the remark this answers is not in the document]
robert-ancell: Argh, I switched from MOCK_METHOD to MOCK_METHODn as that seemed to be the dominant method :)

include/linux_dmabuf_v1.h
```diff
@@ -0,0 +1,68 @@
...
+    LinuxDmabufV1(LinuxDmabufV1 const&) = delete;
+    LinuxDmabufV1& operator=(LinuxDmabufV1 const&) = delete;
+    ~LinuxDmabufV1();
+
+    std::shared_ptr<LinuxDmabufFeedbackV1> get_default_feedback();
+
+private:
+    struct Impl;
+    std::unique_ptr<Impl> const impl;
+};
```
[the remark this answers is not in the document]
robert-ancell: I just had the feedback object first as it had the listener. It felt a bit wonky mixing the two styles so I added this class. I'll change it back though, as it did feel like we should use the Wayland APIs as direct as possible.

tests/linux_dmabuf_v1.cpp
```diff
@@ -0,0 +1,47 @@
...
+{
+};
+
+TEST_F(LinuxDmabufTest, default_feedback)
+{
+    wlcs::Client client{the_server()};
+
+    auto linux_dmabuf = client.bind_if_supported<zwp_linux_dmabuf_v1>(AnyVersion);
+    auto f = wrap_wl_object(zwp_linux_dmabuf_v1_get_default_feedback(linux_dmabuf));
+    auto feedback = std::make_shared<LinuxDmabufFeedbackV1>(f);
```
RAOF, line 37: I think this is going to try to double-free the `default_feedback` object - `wrap_wl_object` is going to put it in a `WlHandle`, and then `LinuxDmabufFeedbackV1` stores the pointer in a `WlHandle` internally, too?

```suggestion
    LinuxDmabufFeedbackV1 feedback{zwp_linux_dmabuf_v1_get_default_feedback(linux_dmabuf)};
```

tests/linux_dmabuf_v1.cpp
```diff
@@ -0,0 +1,47 @@
...
+class LinuxDmabufTest
+    : public wlcs::StartedInProcessServer
+{
+};
+
+TEST_F(LinuxDmabufTest, default_feedback)
+{
+    wlcs::Client client{the_server()};
+
+    auto linux_dmabuf = client.bind_if_supported<zwp_linux_dmabuf_v1>(AnyVersion);
```
RAOF, line 35: This should be 
```suggestion
    auto linux_dmabuf = client.bind_if_supported<zwp_linux_dmabuf_v1>(AtLeastVersion{4});
```
as the feedback interfaces only appear in V4 and above.

tests/linux_dmabuf_v1.cpp
```diff
@@ -0,0 +1,47 @@
...
+    auto f = wrap_wl_object(zwp_linux_dmabuf_v1_get_default_feedback(linux_dmabuf));
+    auto feedback = std::make_shared<LinuxDmabufFeedbackV1>(f);
+
+    EXPECT_CALL(*feedback, format_table(_, _));
+    EXPECT_CALL(*feedback, main_device(_));
+    EXPECT_CALL(*feedback, tranche_target_device(_));
+    EXPECT_CALL(*feedback, tranche_flags(_));
+    EXPECT_CALL(*feedback, tranche_formats(_));
+    EXPECT_CALL(*feedback, tranche_done());
+    EXPECT_CALL(*feedback, done());
```
RAOF, line 45: I guess a *slightly* better test would be to have an `InSequence` to expect that the `done` call is *after* all the previous ones (likewise the `tranch_done`). Since we're poking at other things, maybe that could be rolled in?

RAOF (changes requested):
Still a couple of feedbacks; the last one is optional (but would be nice :wink: )

tests/linux_dmabuf_v1.cpp
robert-ancell: The dones have to be last, but the rest of it is more complex. There could be multiple tranches and the devices/tables/flags/formats could be done in a different order. Looking at the existing `InSequence` code in WLCS and the GMock docs it's not clear how to express this. I think starting with the looser requirements is safest, I'll add a FIXME to improve this.

RAOF (approved):
:+1:

PR #362, opened 2025-08-05
Release 1.8

Saviq:


AlanGriffiths (approved):
Fine. But the "Fixme" comment should probably mention it is armhf specific - otherwise someone will one day try to fix it and...

PR #368, opened 2025-08-07
wlcs: work around a gcc-15 bug

Saviq:


tarek-y-ismail:
Is this a known bug? Might be a good idea to insert a link so we can occasionally check if the bug was fixed

Saviq:
> if it exists

I have not found one, and couldn't get down to a useful, minimal reproducer.

PR #371, opened 2025-09-30
ext-data-control-v1 test draft

tarek-y-ismail:
Adds tests for
* Ensuring copied data arrives at clipboard correctly
TODO:
* Ensuring pasted data arrives from clipboard correctly
* Copying from a client and closing it, clipboard should still hold its data and it should be pastable
* The various error conditions of the protocol


tests/ext_data_control_v1.cpp
```diff
@@ -0,0 +1,338 @@
...
+WLCS_CREATE_INTERFACE_DESCRIPTOR(ext_data_control_manager_v1);
+WLCS_CREATE_INTERFACE_DESCRIPTOR(ext_data_control_source_v1);
+WLCS_CREATE_INTERFACE_DESCRIPTOR(ext_data_control_device_v1);
+WLCS_CREATE_INTERFACE_DESCRIPTOR(ext_data_control_offer_v1);
+}
+
+namespace
+{
+static auto const test_message = "Hello from the other side";
+static auto const test_mime_type = "text/plain";
```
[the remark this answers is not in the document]
tarek-y-ismail: We don't actually get the length of these strings anywhere in the code. Maybe I'm missing something?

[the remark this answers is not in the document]
RAOF: Ah, you used to use them (or, rather, have buffers sized to fit them), when you were sending data over pipes.

tests/ext_data_control_v1.cpp
```diff
@@ -0,0 +1,354 @@
...
+    MockDataOfferListener mdol;
+    EXPECT_CALL(sink.listener, data_offer(_,_))
+        .WillOnce(Invoke([&](struct wl_data_device*, struct wl_data_offer* id)
+        { mdol.listen_to(id); }));
+
+    EXPECT_CALL(mdol, offer(_, _))
+        .WillOnce(Invoke(
+            [&](struct wl_data_offer* offer, char const* mime)
+            {
+                wl_data_offer_receive(offer, mime, test_source_fd);
```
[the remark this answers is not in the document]
tarek-y-ismail: It passes. I suspect this works only because I don't actually write data into the socket.

tests/ext_data_control_v1.cpp
```diff
@@ -0,0 +1,584 @@
...
+    }
+
+    int const &source = pipe_ends[0], sink = pipe_ends[1];
+};
+
+struct DataControlOfferWrapper
+{
+    WlHandle<ext_data_control_offer_v1> const offer;
+    std::vector<std::string> mime_types;
+    bool is_primary{false};
```
tarek-y-ismail: This is only set, but not read from

tests/ext_data_control_v1.cpp
```diff
@@ -0,0 +1,584 @@
...
+    normal,
+    primary
+};
+
+struct ExtDataControlClient: public Client
+{
+    struct DataDeviceState
+    {
+        std::unique_ptr<DataControlOfferWrapper> current_offer;
+    } data_device_state;
```
tarek-y-ismail: Unused

tests/ext_data_control_v1.cpp
```diff
@@ -0,0 +1,584 @@
...
+        auto* self = reinterpret_cast<ExtDataControlClient*>(data);
+
+        if (id == nullptr)
+            return;
+
+        EXPECT_THAT(self->current_offer->offer, Eq(id));
+        self->current_offer->is_primary = false;
+
+        ext_data_control_offer_v1_receive(
+            self->current_offer->offer, self->current_offer->mime_types[0].c_str(), self->receiving_pipe->source);
```
tarek-y-ismail: Should check if we even have any mime types

tests/ext_data_control_v1.cpp
```diff
@@ -0,0 +1,584 @@
...
+    clipboard_client.roundtrip();
+
+    // To receive the `send` event
+    copying_client.roundtrip();
+}
+
+TEST_F(ExtDataControlV1Test, client_copies_into_clipboard_that_started_before_it)
+{
+    ExtDataControlClient clipboard_client{the_server()};
+    clipboard_client.as_sink();
```
tarek-y-ismail: Clipboard should roundtrip  once before the other client to simulate it being already open when the other client starts

tests/ext_data_control_v1.cpp
```diff
@@ -0,0 +1,584 @@
...
+    paste_client.roundtrip();
+
+    // The clipboard should get the `send()` call
+    // from the server and write into the fd
+    clipboard.roundtrip();
+}
+
+TEST_F(ExtDataControlV1Test, setting_the_same_selection_twice_is_a_protocol_error)
+{
+    auto clipboard_client = ExtDataControlClient{the_server()}; // We already set the selection in the ctor
```
tarek-y-ismail: Comment out of date

tests/ext_data_control_v1.cpp
```diff
@@ -0,0 +1,584 @@
...
+
+    CCnPSink sink{the_server()};
+
+    // Client should be able to paste once it gets `selection`
+    EXPECT_CALL(sink.listener, selection(_, _));
+
+    auto f = sink.create_surface_with_focus(); // To get focus
+    sink.roundtrip();
+
+    EXPECT_CALL(clipboard, wrote_data());
```
tarek-y-ismail: `selection` should be strictly after `data_offer` and `offer`. `wrote_data` should be moved below all so the order of the code is similar to the order of operations

tests/ext_data_control_v1.cpp
```diff
@@ -0,0 +1,584 @@
...
+    auto source_client = Client{the_server()};
+    auto source_device_manager = source_client.bind_if_supported<zwp_primary_selection_device_manager_v1>(AnyVersion);
+    auto source_device = PrimarySelectionDevice{source_device_manager, source_client.seat()};
+    auto source_source = PrimarySelectionSource{source_device_manager};
+
+    auto clipboard = ExtDataControlClient{the_server()};
+    clipboard.as_sink();
+
+    // make offer
+    zwp_primary_selection_source_v1_offer(source_source, test_mime_type);
```
tarek-y-ismail: `source.data_offer` should be called before `offer.offer`, which is also called on the incorrect object here.

tests/ext_data_control_v1.cpp
```diff
@@ -0,0 +1,584 @@
...
+
+    EXPECT_CALL(mdol, offer(_, _))
+        .WillOnce(Invoke(
+            [&](struct wl_data_offer* offer, char const* mime)
+            {
+                wl_data_offer_receive(offer, mime, bogus_source_fd);
+
+                clipboard.roundtrip();
+                sink.roundtrip();
+            }));
```
tarek-y-ismail: Should use an `InSequence` object

tests/ext_data_control_v1.cpp
tarek-y-ismail: I think I got confused, `data_offer` is called by the server once `offer` and `set_selection` are called on the client

include/cut_copy_paste.h
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: Christopher James Halse Rogers <christopher.halse.rogers@canonical.com>
```
jhenstridge: This looks to be moved code that had been touched by multiple people. Is this authorship accurate?

src/data_device.cpp
```diff
@@ -21,6 +21,39 @@
...
+    static wl_data_source_listener constexpr listener{
+        .target =
+            [](auto...)
+        {
+        },
+        .send =
+            [](auto* data, auto, auto* mime_type, auto fd)
+        {
+            auto* self = static_cast<DataSource*>(data);
+            self->send_done(mime_type, fd);
```
AlanGriffiths:
```suggestion
        .send =
            [](auto* data, auto... args)
        {
            static_cast<DataSource*>(data)->send(args...);
```

AlanGriffiths (changes requested):
First batch of nits

include/cut_copy_paste.h
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright © 2017 Canonical Ltd.
```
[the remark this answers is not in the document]
tarek-y-ismail: Should it be the same copyright year of the original file, or the year this file was created in? I went with the former since that makes more since to me, but double-checking wouldn't hurt.

tests/ext_data_control_v1.cpp
```diff
@@ -0,0 +1,600 @@
...
+
+struct ExtDataControlV1Test : public wlcs::StartedInProcessServer
+{
+};
+}
+
+TEST_F(ExtDataControlV1Test, client_copies_into_clipboard_that_started_after_it)
+{
+    auto copying_client = ExtDataControlClient{the_server()};
+    auto clipboard_client = ExtDataControlClient{the_server()};
```
AlanGriffiths: This isn't idiomatic C++

```suggestion
    ExtDataControlClient copying_client{the_server()};
    ExtDataControlClient clipboard_client{the_server()};
```

tests/ext_data_control_v1.cpp
```diff
@@ -0,0 +1,600 @@
...
+        case SelectionType::primary:
+            ext_data_control_device_v1_set_primary_selection(data_control_device, source_data_control_source);
+            break;
+        }
+    }
+};
+
+struct ExtDataControlV1Test : public wlcs::StartedInProcessServer
+{
+};
```
AlanGriffiths: All these tests have at least one `ExtDataControlClient` - that can be moved into the fixture
```suggestion
struct ExtDataControlV1Test : public wlcs::StartedInProcessServer
{
    ExtDataControlClient clipboard{the_server()};
};
```

AlanGriffiths (changes requested):

tests/ext_data_control_v1.cpp
```diff
@@ -0,0 +1,638 @@
...
+    }
+
+
+    static void data_control_selection(
+        void* data, struct ext_data_control_device_v1*, struct ext_data_control_offer_v1* id)
+    {
+        auto* self = static_cast<ExtDataControlClient*>(data);
+
+        if (id == nullptr)
+            return;
```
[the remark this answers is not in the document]
tarek-y-ismail: AFAIU, the offer could be null if the clipboard is cleared.

PR #377, opened 2025-11-06
Add tests for `ext_foreign_toplevel_list_v1`

jhenstridge:
Add tests for the [`ext_foreign_toplevel_list_v1` extension](https://wayland.app/protocols/ext-foreign-toplevel-list-v1). This complements https://github.com/canonical/mir/pull/4343, which implements it in Mir.

I used the `wlr_foreign_toplevel_management_unstable_v1` tests as a starting point. Some features not present in the other protocol that are tested include:

1. Handles have an identifier stable across different `ext_foreign_toplevel_list_v1` instances, including instances from different connections.
2. Handles are not destroyed by the server when the corresponding toplevel is unmapped.
3. Handles can be destroyed by the client, even while the corresponding toplevel is still mapped.


tarek-y-ismail (changes requested):
Some nits

PR #378, opened 2025-11-14
spread: refresh Fedora, drop oracular

Saviq:


mattkae (approved):
Cool beans ☕

mattkae:
Although maybe the fedora image is just called `rawhide`?

PR #386, opened 2025-11-19
spread: actually run wlcs, and add debug tmate

Saviq:
The integration .so got split out into a separate package…

### DIFF
--- a/.github/workflows/spread.yml
+++ b/.github/workflows/spread.yml
@@ -57,4 +57,10 @@ jobs:
       uses: actions/checkout@v5
 
     - name: Run Spread task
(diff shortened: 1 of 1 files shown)


.github/workflows/spread.yml
```diff
@@ -57,4 +57,10 @@ jobs:
       uses: actions/checkout@v5
 
     - name: Run Spread task
-      run: snap run spread-mir-ci.spread -v ${{ matrix.spread-task }}
+      run: snap run spread-mir-ci.spread -reuse -v ${{ matrix.spread-task }}
```
tarek-y-ismail: What's the `-reuse` flag for?

PR #387, opened 2025-11-20
Relax ordering expecations in bad buffer test

RAOF:
We expect submitting a bad SHM buffer to result in a protocol exception, but
it's not required to do so before firing the frame callback. Indeed, there
*are* no ordering expectations on when to expect the protocol error.

So, instead of waiting for the frame callback, just dispatch until we
*do* get a protocol exception (or timeout).


AlanGriffiths (approved):
> Looks sensible. If you have the time, split the formatting changes and the actual fix into two PRs?

Having two commits is easy enough for review

PR #399, opened 2025-12-04
Bump version to 1.8.2

AlanGriffiths:
We want to advertise that we have changes not released in 1.8.1. So bump the version on main


RAOF:
> You'll still get 1.8.1 packages, because that's in `debian/changelog` and the latest tag. We need to re-haul the release process for this to work.

Maybe, but this *will* put 1.8.2 in the pkgconf file, which is where we need it.

I'm happy to land this, and (optionally) overhaul our release process later. As it is, this will unblock https://github.com/canonical/mir/pull/4331

RAOF (approved):

PR #431, opened 2026-06-08
Add comprehensive wl_shm conformance tests

robert-ancell:
The existing wl_shm coverage in test_bad_buffer.cpp only exercised the INVALID_FD and INVALID_STRIDE error paths. Add tests/wl_shm.cpp covering the remaining protocol behaviour: mandatory ARGB8888/XRGB8888 format advertisement, the INVALID_FORMAT error, out-of-bounds buffer offsets, pool resizing, and multiple buffers sharing a pool.

### DIFF
--- a/tests/wl_shm.cpp
+++ b/tests/wl_shm.cpp
@@ -106,8 +106,9 @@ TEST_F(ShmTest, create_buffer_exceeding_pool_size_is_an_error)
     EXPECT_PROTOCOL_ERROR(
         {
             // Offsetting by a full row pushes the buffer past the end of the pool.
+            int const offset = stride;
             wl_shm_pool_create_buffer(
-                pool, stride, width, height, stride, WL_SHM_FORMAT_ARGB8888);
+                pool, offset, width, height, stride, WL_SHM_FORMAT_ARGB8888);
             client.roundtrip();
         },
         &wl_shm_pool_interface,
--- a/tests/wl_shm.cpp
+++ b/tests/wl_shm.cpp
@@ -70,7 +70,8 @@ TEST_F(ShmTest, advertises_mandatory_formats)
 // Requesting a buffer with a format the server never advertised must be rejected.
 TEST_F(ShmTest, create_buffer_with_unknown_format_is_an_error)
 {
-    int const width = 4, height = 4;
+    int const width = 4;
+    int const height = 4;
     int const stride = width * 4;
     int const size = stride * height;
     auto const bogus_format = 0xDEADBEEF;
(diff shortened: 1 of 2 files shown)


tests/wl_shm.cpp
```diff
@@ -0,0 +1,176 @@
...
+    auto const formats = collect_advertised_formats(client);
+
+    EXPECT_THAT(formats, Contains(Eq(WL_SHM_FORMAT_ARGB8888)));
+    EXPECT_THAT(formats, Contains(Eq(WL_SHM_FORMAT_XRGB8888)));
+}
+
+// Requesting a buffer with a format the server never advertised must be rejected.
+TEST_F(ShmTest, create_buffer_with_unknown_format_is_an_error)
+{
+    int const width = 4, height = 4;
```
mattkae:
```suggestion
    int const width = 4;
    int const height = 4;
```

tests/wl_shm.cpp
```diff
@@ -0,0 +1,176 @@
...
+    int const stride = width * 4;
+    int const size = stride * height;
+    auto const bogus_format = 0xDEADBEEF;
+
+    auto fd = wlcs::helpers::create_anonymous_file(size);
+    auto pool = wl_shm_create_pool(client.shm(), fd, size);
```
mattkae:
```suggestion
    auto const fd = wlcs::helpers::create_anonymous_file(size);
    auto const pool = wl_shm_create_pool(client.shm(), fd, size);
```

tests/wl_shm.cpp
mattkae: And below

tests/wl_shm.cpp
mattkae: And below

tests/wl_shm.cpp
```diff
@@ -0,0 +1,176 @@
...
+    close(fd);
+
+    EXPECT_PROTOCOL_ERROR(
+        {
+            // Offsetting by a full row pushes the buffer past the end of the pool.
+            wl_shm_pool_create_buffer(
+                pool, stride, width, height, stride, WL_SHM_FORMAT_ARGB8888);
```
mattkae: Nit for readability:

```suggestion
            // Offsetting by a full row pushes the buffer past the end of the pool.
            int const offset = stride;
            wl_shm_pool_create_buffer(
                pool, offset, width, height, stride, WL_SHM_FORMAT_ARGB8888);
```

mattkae (approved):
I have some nits, but looks good overall.

PR #432, opened 2026-06-09
Add some basic tests for toplevel image capture

jhenstridge:
This adds some basic tests for image capture from toplevels:

* Capture sources can be created from toplevels
* Capture sessions can be created from the above source.
* Capturing a frame from the session returns data that looks like it comes from the surface's buffer.
* A new frame is offered for capture when a new buffer is committed to the surface.
* The session reports buffer constraints changes when the surface is resized.
* Frames can be captured when the toplevel is minimised.
* The capture session sends the stopped event when it's corresponding surface is destroyed.

Related: https://github.com/canonical/mir/pull/5002

### DIFF
--- a/include/in_process_server.h
+++ b/include/in_process_server.h
@@ -202,6 +202,7 @@ private:
 };
 
 class Client;
+class ShmBuffer;
 
 class Surface
 {
@@ -214,8 +215,10 @@ public:
     operator ::wl_surface*() const;
     auto wl_surface() const -> ::wl_surface* { return *this; };
 
+    void attach_buffer(ShmBuffer const& buffer);
     void attach_buffer(int width, int height);
     void add_frame_callback(std::function<void(int)> const& on_frame);
+    void attach_visible_buffer(ShmBuffer const& buffer);
     void attach_visible_buffer(int width, int height);
     void run_on_destruction(std::function<void()> callback);
 
--- a/src/ext_foreign_toplevel_list_v1.cpp
+++ b/src/ext_foreign_toplevel_list_v1.cpp
@@ -192,7 +192,7 @@ auto wlcs::ExtForeignToplevelListV1::toplevel(std::string const& app_id) const -
         }
     }
     if (!match)
-        BOOST_THROW_EXCEPTION(std::runtime_error("No toplevels have the app ID " + app_id));
+        BOOST_THROW_EXCEPTION(std::out_of_range("No toplevels have the app ID " + app_id));
     if (match.value()->is_dirty())
         BOOST_THROW_EXCEPTION(std::runtime_error("Toplevel has pending updates"));
     return *match.value();
--- a/src/ext_foreign_toplevel_list_v1.cpp
+++ b/src/ext_foreign_toplevel_list_v1.cpp
@@ -1,0 +1,208 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "ext_foreign_toplevel_list_v1.h"
+#include "in_process_server.h"
+#include "version_specifier.h"
+#include "wl_handle.h"
+
+#include <boost/throw_exception.hpp>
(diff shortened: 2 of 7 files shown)


src/ext_foreign_toplevel_list_v1.cpp
```diff
@@ -0,0 +1,208 @@
...
+        if (i->app_id() == app_id)
+        {
+            if (match)
+                BOOST_THROW_EXCEPTION(std::runtime_error("Multiple toplevels have the same app ID " + app_id));
+            else
+                match = i.get();
+        }
+    }
+    if (!match)
+        BOOST_THROW_EXCEPTION(std::runtime_error("No toplevels have the app ID " + app_id));
```
RAOF: Minor nit, this could be `std::out_of_range`.

(Edit: I see this is copied from elsewhere; still, while we're here…)

tests/ext_image_copy_capture_v1.cpp
```diff
@@ -580,4 +612,225 @@ TEST_F(ExtImageCopyCaptureTest, cursor_session_captures_pointer_image)
...
+
+    auto source = wlcs::wrap_wl_object(ext_foreign_toplevel_image_capture_source_manager_v1_create_source(source_manager, list.toplevel()));
+    ImageCopyCaptureSession session{ext_image_copy_capture_manager_v1_create_session(capture_manager, source, 0)};
+    client.roundtrip();
+
+    ASSERT_THAT(session.is_dirty(), IsFalse());
+    ASSERT_THAT(session.buffer_size(), Ne(std::nullopt));
+    auto buffer_size = session.buffer_size().value();
+    // TODO: compositor could report different formats
+    ASSERT_THAT(session.shm_formats(), Contains(Eq(WL_SHM_FORMAT_ARGB8888)));
```
RAOF, line 623: Non-blocking, but you *could* turn these into skipped tests rather than failures by doing something like `wlcs::ExtensionExpectedlyNotSupported` does — record a `wlcs-skip-test` property and bail with an exception.

tests/ext_image_copy_capture_v1.cpp
```diff
@@ -580,4 +612,225 @@ TEST_F(ExtImageCopyCaptureTest, cursor_session_captures_pointer_image)
...
+
+    // New buffer constraints are sent after the toplevel is resized
+    surface.attach_visible_buffer(200, 200);
+    client.dispatch_until([&session, buffer_size]() { return !session.is_dirty() && session.buffer_size() != buffer_size; });
+    EXPECT_THAT(session.buffer_size(), Eq(wlcs::Size{buffer_size.width*2, buffer_size.height*2}));
```
RAOF, line 808: Is this not `wlcs::Size{200, 200}` because we don't know if scaling is in play?

include/in_process_server.h
```diff
@@ -190,8 +191,10 @@ class Surface
     operator ::wl_surface*() const;
     auto wl_surface() const -> ::wl_surface* { return *this; };
 
+    void attach_buffer(ShmBuffer const& buffer);
     void attach_buffer(int width, int height);
     void add_frame_callback(std::function<void(int)> const& on_frame);
+    void attach_visible_buffer(ShmBuffer const& buffer);
     void attach_visible_buffer(int width, int height);
```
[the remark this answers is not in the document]
RAOF: Copilot is wrong here; we don't care if you pass a temporary, because the buffer *only* needs to be valid at the time of the `wl_surface_attach()` call; it's safe to destroy it immediately after.

tests/ext_image_copy_capture_v1.cpp
```diff
@@ -580,4 +612,225 @@ TEST_F(ExtImageCopyCaptureTest, cursor_session_captures_pointer_image)
...
+    wlcs::ExtForeignToplevelListV1 list{client};
+    auto surface = client.create_visible_surface(100, 100);
+
+    // Attach our own buffer
+    wlcs::ShmBuffer surface_buffer{client, 100, 100};
+    memset(surface_buffer.data().data(), 0x7f, surface_buffer.data().size());
+    surface.attach_visible_buffer(surface_buffer);
+
+    auto source = wlcs::wrap_wl_object(ext_foreign_toplevel_image_capture_source_manager_v1_create_source(source_manager, list.toplevel()));
+    ImageCopyCaptureSession session{ext_image_copy_capture_manager_v1_create_session(capture_manager, source, 0)};
```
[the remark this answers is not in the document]
RAOF: Copilot is wrong-ish here. `attach_visible_buffer()` does a dispatch loop until we get a frame event, which will be at least as good as `client.roundtrip()` here.

However, I don't think there's a guarantee that the toplevel handle must have been sent by then, so it probably *would* be more reliable to do a `client.dispatch_until(…list.toplevel() resolves, probably needing to do a try {} catch dance…)`.

RAOF (changes requested):
Looks good; enough nits that I've not just approved outright, but once they're settled it'll be good to go.

RAOF (approved):
Yup! That looks good now.

PR #433, opened 2026-06-09
Add wl_shm create_pool fd and size validation tests

robert-ancell:
Add tests verifying that wl_shm.create_pool rejects an unmappable file descriptor (a pipe) with INVALID_FD, and a non-positive size with INVALID_STRIDE.


mattkae (approved):
Very legit !

PR #434, opened 2026-06-09
Add wl_registry conformance tests

robert-ancell:
Add tests/wl_registry.cpp verifying that wl_display.get_registry returns the server's globals and that a registry can be destroyed using wl_fixes.destroy_registry.

### DIFF
--- a/tests/wl_registry.cpp
+++ b/tests/wl_registry.cpp
@@ -105,29 +105,20 @@ TEST_F(RegistryTest, registry_can_be_destroyed_with_wl_fixes)
     // Destroy a separate registry via wl_fixes; the client must not use it after.
     auto doomed_registry = wl_display_get_registry(client);
     client.roundtrip();
-
     auto const doomed_id = wl_proxy_get_id(reinterpret_cast<wl_proxy*>(doomed_registry));
-
     wl_fixes_destroy_registry(fixes, doomed_registry);
-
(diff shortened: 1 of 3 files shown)


tests/wl_registry.cpp
```diff
@@ -0,0 +1,124 @@
...
+    auto doomed_registry = wl_display_get_registry(client);
+    client.roundtrip();
+    auto const doomed_id = wl_proxy_get_id(reinterpret_cast<wl_proxy*>(doomed_registry));
+    wl_fixes_destroy_registry(fixes, doomed_registry);
+    client.roundtrip();
+
+    // The compositor must emit wl_display.delete_id for the destroyed registry.
+    // We cannot observe delete_id directly (libwayland handles it internally), but
+    // a freed ID is re-used by the next object the client creates. If delete_id had
+    // not been emitted, the ID would remain reserved and a fresh ID would be used.
```
mattkae, line 115: That is wacky :)

mattkae (approved):
I don't love it, but I can accept it

tests/wl_registry.cpp
robert-ancell: Yeah, it's definitely quirky. I did try to see if there was a better alternative but I couldn't find it. I think the only alternative was either to modify libwayland or have our own client implementation in wlcs.

PR #435, opened 2026-06-10
Update wayland.xml and handle new core protocol events

robert-ancell:
Bump the bundled wayland.xml to a newer revision and add handlers for the new events that WLCS now binds to:

- wl_output v4: name and description events
- wl_surface v6: preferred_buffer_scale and preferred_buffer_transform events

Without these, the positionally-initialized listener structs would leave null function pointers, crashing when compositors emit the events.

### DIFF
--- a/src/in_process_server.cpp
+++ b/src/in_process_server.cpp
@@ -1295,11 +1295,21 @@ public:
             me->pending.scale = factor;
         }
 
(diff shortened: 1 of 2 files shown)


jhenstridge (approved):
Looks reasonable.

PR #437, opened 2026-06-12
Add xdg_surface configure and ack_configure tests

robert-ancell:
Add tests covering xdg_surface.ack_configure error handling and configure event behaviour:
- ack_configure with an invalid (never-sent) serial is an error
- ack_configure twice for the same configure event is an error
- ack_configure without any configure event is an error
- ack_configure for an earlier configure after a later one is an error
- multiple configure events are received on state changes
- only the last of multiple configure events needs acknowledging

### DIFF
--- a/tests/xdg_surface_stable.cpp
+++ b/tests/xdg_surface_stable.cpp
@@ -33,6 +33,8 @@
 
 #include <gmock/gmock.h>
 
+#include <vector>
+
 using namespace testing;
 using XdgSurfaceStableTest = wlcs::InProcessServer;
 using wlcs::AnyVersion;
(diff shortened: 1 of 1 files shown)


tests/xdg_surface_stable.cpp
```diff
@@ -66,6 +68,269 @@ TEST_F(XdgSurfaceStableTest, gets_configure_event)
...
+    };
+
+    // Map the surface, acking the initial configure as required.
+    wait_for_another_configure();
+    surface.attach_buffer(200, 320);
+    wl_surface_commit(surface);
+
+    // Stop acking so we can accumulate several un-acked configure events.
+    auto_ack = false;
+    auto const first_reconfigure_index = configure_serials.size();
```
tarek-y-ismail:
```suggestion
    auto const first_reconfigure_index = configure_serials.size() - 1;
```

tests/xdg_surface_stable.cpp
```diff
@@ -66,6 +68,269 @@ TEST_F(XdgSurfaceStableTest, gets_configure_event)
...
+        client.roundtrip();
+    }
+    catch(wlcs::ProtocolError const& error)
+    {
+        EXPECT_THAT(error.interface(), Eq(&xdg_surface_interface));
+        EXPECT_THAT(error.error_code(), Eq(XDG_SURFACE_ERROR_INVALID_SERIAL));
+        return;
+    }
+
+    FAIL() << "Expected protocol error not received";
```
tarek-y-ismail: Could be replaced with `EXPECT_PROTOCOL_ERROR`

tests/xdg_surface_stable.cpp
```diff
@@ -66,6 +68,269 @@ TEST_F(XdgSurfaceStableTest, gets_configure_event)
...
```
tarek-y-ismail: Again, `EXPECT_PROTOCOL_ERROR`

tests/xdg_surface_stable.cpp
```diff
@@ -66,6 +68,269 @@ TEST_F(XdgSurfaceStableTest, gets_configure_event)
...
```
tarek-y-ismail: One more!

tests/xdg_surface_stable.cpp
```diff
@@ -66,6 +68,269 @@ TEST_F(XdgSurfaceStableTest, gets_configure_event)
...
```
tarek-y-ismail: And another!

tests/xdg_surface_stable.cpp
```diff
@@ -66,6 +68,269 @@ TEST_F(XdgSurfaceStableTest, gets_configure_event)
...
+    // Wait for the initial configure, ack it, and map the surface with a buffer.
+    wait_for_another_configure();
+    surface.attach_buffer(200, 320);
+    wl_surface_commit(surface);
+
+    // Requesting a state change causes the compositor to send a further configure.
+    xdg_toplevel_set_maximized(toplevel);
+    wait_for_another_configure();
+
+    EXPECT_THAT(configure_serials.size(), Ge(2u));
```
tarek-y-ismail: `dispatch_until` will throw an exception if its condition is not met before the timeout. Which means that reaching this line means that the two calls to `wait_for_another_configure` must have succeeded, and the number of configures must be greater than or equal 2. So this check is redundant.

tarek-y-ismail (changes requested):
Looks good, just need to clean up the code a bit. In general, there's a lot of redundancy that I think could be reduced to a reasonable amount.

tests/xdg_surface_stable.cpp
tarek-y-ismail: So  I guess this would fit better as `size - 1` right after maximization and waiting for `configure` instead of here?  Both should be valid but seeing `size()` stored as an index sets of my spidey senses.

tarek-y-ismail (approved):

PR #438, opened 2026-06-14
Use EXPECT_PROTOCOL_ERROR in tests

robert-ancell:
Convert try/catch wlcs::ProtocolError blocks to the EXPECT_PROTOCOL_ERROR macro in:

- tests/ext_image_copy_capture_v1.cpp
- tests/test_bad_buffer.cpp
- tests/wlr_layer_shell_v1.cpp
- tests/xdg_surface_stable.cpp

### DIFF
--- a/tests/test_bad_buffer.cpp
+++ b/tests/test_bad_buffer.cpp
@@ -31,6 +31,7 @@
 #include "helpers.h"
 
 #include "in_process_server.h"
+#include "expect_protocol_error.h"
 
 /* tests, that attempt to crash the compositor on purpose */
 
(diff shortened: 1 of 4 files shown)


robert-ancell:
See also https://github.com/canonical/wlcs/pull/439 - in this case the specific error code was not being checked, so I made it a separate PR

tests/test_bad_buffer.cpp
```diff
@@ -82,20 +83,12 @@ TEST_F(BadBufferTest, test_truncated_shm_file)
...
-        wl_buffer_destroy(bad_buffer);
-        EXPECT_THAT(err.error_code(), Eq(WL_SHM_ERROR_INVALID_FD));
-        EXPECT_THAT(err.interface(), Eq(&wl_buffer_interface));
-        return;
-    }
-
-    FAIL() << "Expected protocol error not raised";
+    }, &wl_buffer_interface, WL_SHM_ERROR_INVALID_FD);
+
+    wl_buffer_destroy(bad_buffer);
```
RAOF, line 91: Hm. This (existing) is kinda wrong; once we've hit a protocol error the display is not gonna do anything useful. It's also not going to be a *problem*, so 🤷‍♀️

RAOF (approved):
Yeah, reasonable.

Note that this does change the behaviour of some of the tests - `EXPECT_PROTOCOL_ERROR` doesn't bail out early. Probably it should! But that doesn't make any of these tests incorrect.

PR #440, opened 2026-06-15
Add drag-and-drop tests for wl_data_device

robert-ancell:
Add tests/data_device_drag_and_drop.cpp covering the previously untested wl_data_device.start_drag flow: starting a drag, and the data_offer, enter, motion, leave and drop events delivered to the data device, plus the offered mime type on the wl_data_offer.


tests/data_device_drag_and_drop.cpp
```diff
@@ -0,0 +1,195 @@
...
+    }
+};
+}
+
+TEST_F(DragAndDrop, data_device_receives_offer_when_drag_enters_target_surface)
+{
+    EXPECT_CALL(device_listener, data_offer(_, _)).Times(AtLeast(1));
+
+    start_drag();
+    drag_over_target();
```
RAOF: So, as a matter of style I broadly think that what the test cases are testing should be immediately visible by reading the test case. There's obviously a balance here between de-duplicating setup and explicitness of the test, but I think this is clearly too far in the former direction :).

Perhaps `start_drag(x, y)` should just do the mouse event bit, returning the serial?

tests/data_device_drag_and_drop.cpp
```diff
@@ -0,0 +1,195 @@
...
+    DataDevice data_device{wl_data_device_manager_get_data_device(manager, client.seat())};
+    NiceMock<MockDataDeviceListener> device_listener{data_device};
+    DataSource data_source{wl_data_device_manager_create_data_source(manager)};
+
+    Surface source_surface{client.create_visible_surface(surface_width, surface_height)};
+    Surface target_surface{client.create_visible_surface(surface_width, surface_height)};
+
+    wlcs::Pointer pointer = the_server().create_pointer();
+
+    bool entered_target{false};
```
RAOF: This is kinda weird hanging around all by itself here, being used only in `drag_over_target`

RAOF (changes requested):

RAOF (approved):
Yeah, that's better.

PR #441, opened 2026-06-15
Add wl_surface buffer state tests

robert-ancell:
Add tests/wl_surface.cpp covering the previously untested wl_surface buffer-state requests: attach, set_buffer_scale, set_buffer_transform (90/180/270), damage_buffer and offset, along with the invalid_scale, invalid_transform and invalid_offset protocol errors.

The positive tests currently only confirm the requests are accepted and the surface is presented; comments describe how to extend them to verify the state is actually applied.


tests/wl_surface.cpp
```diff
@@ -0,0 +1,210 @@
...
+namespace
+{
+struct SurfaceTest : StartedInProcessServer
+{
+    Client client{the_server()};
+
+    auto compositor_version() const -> uint32_t
+    {
+        return wl_proxy_get_version(reinterpret_cast<wl_proxy*>(client.compositor()));
+    }
```
RAOF: This is a weird re-implementation of `wl_compositor_get_version`?

tests/wl_surface.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+    wl_surface_attach(surface, buffer, 0, 0);
+    commit_and_wait(surface);
+}
+
+TEST_F(SurfaceTest, buffer_scale)
+{
+    if (compositor_version() < 3)
+        GTEST_SKIP() << "wl_compositor version does not support set_buffer_scale";
+
```
RAOF, line 77: This is not the correct way to skip tests based on protocol version. You're looking for 
`auto compositor = client.bind_if_supported<wl_compositor>(AtLeastVersion{3});`

This'll handle the skip, but also handle where servers lie about the version of an interface they support.

This will require adding a `WlInterfaceDescriptor<wl_compositor>` specialisation.

tests/wl_surface.cpp
```diff
@@ -0,0 +1,210 @@
...
+    // damaged region (and only that region) was updated on screen.
+    auto surface = client.create_visible_surface(100, 100);
+    ShmBuffer buffer{client, 100, 100};
+
+    wl_surface_attach(surface, buffer, 0, 0);
+    wl_surface_damage_buffer(surface, 0, 0, 100, 100);
+    commit_and_wait(surface);
+}
+
+TEST_F(SurfaceTest, offset)
```
RAOF, line 148: If you added `AtMostVersion` to `version_specifier.h`, you could also test that `attach` with non-zero offset works!

RAOF (changes requested):

include/in_process_server.h
```diff
@@ -48,6 +48,7 @@ namespace wlcs
 {
 class VersionSpecifier;
 
+WLCS_CREATE_INTERFACE_DESCRIPTOR(wl_compositor)
```
RAOF: We should actually manually specialise this because `wl_compositor` has [gained a destructor in version 7](https://wayland.app/protocols/wayland#wl_compositor:request:release), so we should do the same thing as the `wl_output` specialisation (which similarly gained a destructor).

RAOF (changes requested):
Still a nit with the interface descriptor, but otherwise looks good!

include/in_process_server.h
robert-ancell: That wasn't even in the XML! But I still did it :)

RAOF (approved):
Yup!

PR #445, opened 2026-06-16
Add more comprehensive wl_output tests

robert-ancell:
Replace the two existing wl_output tests with a more thorough set that verifies each property advertised in the initial output advertisement.

A mockable WlOutputListener wraps the wl_output events so the tests can assert the advertised geometry, current mode, scale and name via EXPECT_CALL, and confirm the advertisement is terminated by exactly one done event.


tests/wl_output.cpp
```diff
@@ -22,26 +22,207 @@
...
+    MOCK_METHOD(void, scale, (wl_output*, int32_t), (override));
+    MOCK_METHOD(void, name, (wl_output*, char const*), (override));
+    MOCK_METHOD(void, description, (wl_output*, char const*), (override));
+};
+
+/// True if the mode flags mark it as the output's current mode.
+MATCHER(IsCurrentMode, "")
+{
+    return arg & WL_OUTPUT_MODE_CURRENT;
+}
```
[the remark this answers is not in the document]
robert-ancell: The default behaviour is to generate a description, which will be appropriate

tests/wl_output.cpp
```diff
@@ -22,26 +22,207 @@
...
+        client.dispatch_until([&done]() { return done; });
+    }
+};
+
+TEST_F(WlOutputTest, sends_done_after_initial_properties)
+{
+    wlcs::Client client{the_server()};
+
+    auto const output = client.bind_if_supported<wl_output>(wlcs::AnyVersion);
+    NiceMock<MockWlOutputListener> listener{output};
```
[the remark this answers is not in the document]
robert-ancell: This throws an exception if it would have been nullptr, so is not required.

tests/wl_output.cpp
```diff
@@ -22,26 +22,207 @@
...
-TEST_F(WlOutputTest, wl_output_properties_set)
+struct WlOutputTest : wlcs::InProcessServer
+{
+    /// Binds a fresh wl_output and dispatches until the server has finished
+    /// advertising its properties (i.e. it has sent the initial wl_output.done).
+    void receive_initial_properties(wlcs::Client& client, MockWlOutputListener& listener)
+    {
+        bool done = false;
+        EXPECT_CALL(listener, done(_)).WillRepeatedly(Invoke([&done](auto) { done = true; }));
+        client.dispatch_until([&done]() { return done; });
```
jhenstridge: This `WillRepeatedly` lambda may continue to be invoked after the method returns. The reference to the `done` local will be out of scope at that point, so you're overwriting whatever happens to take its place on the stack.

You probably want a `Mock::VerifyAndClearExpectations` call to clean things up, and perhaps install a new handler if you want to handle further `done` events. Check `create_xdg_shell_stable_surface` for a similar pattern.

tests/wl_output.cpp
```diff
@@ -22,26 +22,207 @@
...
 
-#include <memory>
-
 using namespace testing;
 using namespace wlcs;
 
-using WlOutputTest = wlcs::InProcessServer;
+namespace
+{
+/// Forwards wl_output events to virtual methods so they can be mocked.
```
RAOF: Why did you choose to use inheritance to dispatch here? There doesn't seem to be any benefit over just having the `MOCK_METHOD()`s on the class and have the thunks directly call those; that's what we do in most of these cases (eg: `xdg_shell_stable.h`, `linux_dmabuf_v1.h`, etc)

tests/wl_output.cpp
robert-ancell: I changed it to `ON_CALL`/`WillByDefault` as suggested by Gemini. This is used elsewhere. Do you think this is appropriate?

tests/wl_output.cpp
robert-ancell: Changed to match

tests/wl_output.cpp
```diff
@@ -24,21 +24,128 @@
...
+
+struct WlOutputTest : wlcs::InProcessServer
+{
+    /// Binds a fresh wl_output and dispatches until the server has finished
+    /// advertising its properties (i.e. it has sent the initial wl_output.done).
+    void receive_initial_properties(wlcs::Client& client, WlOutputListener& listener)
+    {
+        bool done = false;
+        ON_CALL(listener, done()).WillByDefault(Invoke([&done]() { done = true; }));
+        client.dispatch_until([&done]() { return done; });
```
jhenstridge, line 76: There should probably be a `VerifyAndClearExpectations` call here to ensure the closure is not called again after the `done` local is out of scope.

I realise that you're only calling this method at the end of test methods, but it would make it more difficult to misuse.

tests/wl_output.cpp
```diff
@@ -24,21 +24,128 @@
...
+}
+
+TEST_F(WlOutputTest, advertises_scale)
+{
+    wlcs::Client client{the_server()};
+
+    auto const output = client.bind_if_supported<wl_output>(wlcs::AnyVersion);
+    NiceMock<WlOutputListener> listener{output};
+
+    EXPECT_CALL(listener, scale(Ge(1)));
```
jhenstridge, line 131: The protocol documentation says this event is optional, so this doesn't seem quite right.

Adding `.Times(AtMost(1))` to this expectation would handle the case of the event not being sent. But because you're using `NiceMock`, any event received with `factor < 1` is going to be ignored due to not matching any expectations. So maybe the following would be better?

```suggestion
    EXPECT_CALL(listener, scale(_)).Times(0);
    EXPECT_CALL(listener, scale(Ge(1))).Times(AtMost(1));
```

The first expectation will fail on any bad scale events, and the second will accept any good scale events.

It might be worth doing the same for the other tests: while they will fail if an event with the expected arguments doesn't arrive, they won't tell you about unexpected events.

tests/wl_output.cpp
```diff
@@ -24,21 +24,128 @@
...
+    EXPECT_CALL(listener, mode(IsCurrentMode(), Gt(0), Gt(0), Ge(0)));
+
+    receive_initial_properties(client, listener);
+}
+
+TEST_F(WlOutputTest, advertises_scale)
+{
+    wlcs::Client client{the_server()};
+
+    auto const output = client.bind_if_supported<wl_output>(wlcs::AnyVersion);
```
jhenstridge, line 128: The `scale` event was added in version 2 of the interface.

```suggestion
    auto const output = client.bind_if_supported<wl_output>(wlcs::AtLeastVersion{WL_OUTPUT_SCALE_SINCE_VERSION});
```

tests/wl_output.cpp
```diff
@@ -24,21 +24,128 @@
...
+        ON_CALL(listener, done()).WillByDefault(Invoke([&done]() { done = true; }));
+        client.dispatch_until([&done]() { return done; });
+    }
+};
+
+TEST_F(WlOutputTest, sends_done_after_initial_properties)
+{
+    wlcs::Client client{the_server()};
+
+    auto const output = client.bind_if_supported<wl_output>(wlcs::AnyVersion);
```
jhenstridge, line 84: The `done` event was added in version 2 of the interface.

```suggestion
    auto const output = client.bind_if_supported<wl_output>(wlcs::AtLeastVersion{WL_OUTPUT_DONE_SINCE_VERSION});
```

In fact, any tests using the `receive_initial_properties` helper will fail without that version, so I don't think `AnyVersion` is appropriate for any of the tests here.

tests/wl_output.cpp
```diff
@@ -24,21 +24,128 @@
...
+}
+
+TEST_F(WlOutputTest, advertises_name)
+{
+    wlcs::Client client{the_server()};
+
+    auto const output = client.bind_if_supported<wl_output>(wlcs::AtLeastVersion{WL_OUTPUT_NAME_SINCE_VERSION});
+    NiceMock<WlOutputListener> listener{output};
+
+    EXPECT_CALL(listener, name(StrNe("")));
```
jhenstridge, line 143: It might also be useful to check that no two outputs return the same name, but I'm not sure how you'd bind the other outputs via `bind_if_supported`.

jhenstridge (changes requested):
There's a few things that need fixing here, but it mostly comes down to two main issues:

1. All the tests that rely on the done event require at least version 2 of the interface.
2. The use of `NiceMock` means unexpected calls will be ignored. It might be worth adding a catch all `EXPECT_CALL(listener, method(_)).Times(0);` expectation to explicitly fail on events with bad arguments.

tests/wl_output.cpp
jhenstridge: I don't think `WillByDefault` makes things any better, since the callback will still be attached to the mock after the captured reference is out of scope.

Maybe this would be the most concise and correct version?

```
EXPECT_CALL(listener, done()).WillOnce(Invoke([&done](auto) { done = true; })).RetiresOnSaturation();
```

That should remove the expectation from the mock after the event is dispatched, so it doesn't last past the end of the function call.

PR #446, opened 2026-06-16
Remove unused includes

robert-ancell:
Drop headers that are no longer referenced by the including file:
- in_process_server.h: shared_library.h
- gtk_primary_selection.h: wl_handle.h, <mutex>, <set>
- wl_output.cpp: helpers.h, <memory>
- xdg_popup.cpp, xdg_surface_stable.cpp, xdg_surface_v6.cpp: helpers.h
- xdg_toplevel_stable.cpp: helpers.h, <optional>


jhenstridge (approved):
Looks good!

PR #452, opened 2026-06-26
Add wl_pointer cursor-role tests

robert-ancell:
Add tests/wl_pointer.cpp covering wl_pointer.set_cursor behaviour: role-less cursor surfaces, hiding the cursor with a null surface (both with and without a previously set cursor), reassigning the cursor role, the protocol errors raised when set_cursor targets a surface that already holds another role (xdg or subsurface) or when a cursor surface is later given a different role, and that a stale serial is ignored.

Note that some wl_pointer enter/leave/motion coverage already exists in tests/wl_surface_events.cpp; those tests may be moved here in the future.

### DIFF
--- a/tests/wl_pointer.cpp
+++ b/tests/wl_pointer.cpp
@@ -1,0 +1,204 @@
+/*
+ * Copyright © 2024 Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "in_process_server.h"
+#include "expect_protocol_error.h"
+
+#include <gmock/gmock.h>
+
+using namespace testing;
+
+namespace
+{
+class WlPointerTest : public wlcs::StartedInProcessServer
+{
+public:
+    WlPointerTest()
+        : client{the_server()},
+          pointer{the_server().create_pointer()},
+          surface{client.create_visible_surface(surface_width, surface_height)}
+    {
+        the_server().move_surface_to(surface, surface_x, surface_y);
+    }
+
+    /// Moves the pointer onto the test surface and returns the serial of the
+    /// wl_pointer.enter event this produces. Many wl_pointer requests, most
+    /// notably set_cursor, must be issued with a valid enter serial.
+    auto move_pointer_to_surface() -> uint32_t
+    {
+        std::optional<uint32_t> enter_serial;
+        client.add_pointer_enter_notification(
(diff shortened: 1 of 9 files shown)


tests/wl_pointer.cpp
```diff
@@ -0,0 +1,208 @@
...
+            {
+                // Inside the enter handler the client's latest serial is, by
+                // definition, the enter serial.
+                enter_serial = client.latest_serial();
+                return false;
+            });
+        pointer.move_to(surface_x + pointer_offset_x, surface_y + pointer_offset_y);
+        client.dispatch_until([&]{ return enter_serial.has_value(); });
+        return enter_serial.value();
+    }
```
jhenstridge, line 51: Would it make sense to split this out into shared code? It looks almost the same as some code in the image-copy-capture tests.

tests/wl_pointer.cpp
```diff
@@ -0,0 +1,208 @@
...
+};
+}
+
+TEST_F(WlPointerTest, set_cursor_with_a_role_less_surface_is_accepted)
+{
+    auto const enter_serial = move_pointer_to_surface();
+
+    auto cursor = make_cursor_surface();
+
+    EXPECT_NO_THROW({
```
jhenstridge: Does wrapping this in EXPECT_NO_THROW add much value here? If it throws an exception, the test will fail.

tests/wl_pointer.cpp
```diff
@@ -0,0 +1,208 @@
...
+}
+
+TEST_F(WlPointerTest, set_cursor_on_a_subsurface_is_a_protocol_error)
+{
+    auto const enter_serial = move_pointer_to_surface();
+
+    // Creating a subsurface gives the wl_surface the subsurface role.
+    wlcs::Subsurface subsurface{surface};
+
+    EXPECT_PROTOCOL_ERROR({
```
jhenstridge: Would it make sense to limit this EXPECT_PROTOCOL_ERROR check to just the `roundtrip` call?

tests/wl_pointer.cpp
```diff
@@ -0,0 +1,208 @@
...
+    });
+}
+
+TEST_F(WlPointerTest, set_cursor_with_null_surface_hides_the_cursor)
+{
+    auto const enter_serial = move_pointer_to_surface();
+
+    EXPECT_NO_THROW({
+        wl_pointer_set_cursor(client.the_pointer(), enter_serial, nullptr, 0, 0);
+        client.roundtrip();
```
jhenstridge: This doesn't seem to actually check the thing that the test claims to verify: just that the server accepts the messages that would cause the cursor to be hidden.

You could probably verify it indirectly via image-copy-capture, or perhaps add a new API to the test harness. It would be relatively easy to provide the current cursor image in Mir. Not sure how easy it'd be for other compositors though.

The same goes for most of the other tests added here.

jhenstridge (commented):
My main concern here is that the WLCS infrastructure doesn't currently have a way to test a bunch of the things these tests claim to (unless you bring image-copy-capture into the picture, where a lot more code would be under test).

There's probably still some value in showing that the server doesn't reject these messages though.

tests/wl_pointer.cpp
robert-ancell: Added comments about these limitations.

RAOF (approved):
This looks reasonable to me, to the extent we have infrastructure for this sort of thing.

PR #453, opened 2026-07-01
Test subtracted input regions

robert-ancell:
Add coverage for wl_region_subtract in the input-region tests: a region covering the whole surface with a rectangular hole subtracted from the middle. New edge and corner test points check that input is seen on the surrounding frame but not inside the subtracted hole, replacing the 'TODO: test subtract' placeholder.


tarek-y-ismail (approved):
Oops. Forgot to approve.

PR #464, opened 2026-09-04
wl_pointer.leave should never have a null surface and wl_pointer.enter should not assert a leave if the surface is invalid

mattkae:
## Problem
Our WLCS test suite assumed that a `wl_pointer.leave` event could fire with a null surface, but this is incorrect. Null must be explicitly allowed per the Wayland protocol, so any event sending null here would be an error.

## Solution
- Disallow null surfaces in wl_pointer.leave. This should be a `FAIL` going forward
- When a surface is destroyed, we invalidate any state that it was associated with, as future events cannot reliably compare against a window that has been destroyed

pairs with https://github.com/canonical/mir/pull/5180

### DIFF
--- a/include/in_process_server.h
+++ b/include/in_process_server.h
@@ -372,6 +372,11 @@ public:
      */
     void flush();
 
+    /**
+     * When a surface is dropped by WLCS, 
+     */
+    void invalidate_surface(wl_surface* surface);
+
 private:
     class Impl;
     std::unique_ptr<Impl> const impl;
--- a/include/in_process_server.h
+++ b/include/in_process_server.h
@@ -372,12 +372,24 @@ public:
      */
     void flush();
 
+private:
+    friend class Surface;
+
     /**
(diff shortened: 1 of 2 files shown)


RAOF (approved):
Hm.

This is reasonable, but I notice that *Mutter* sends a `wl_pointer.leave(<serial>, nil)` event. I think we actually need to discover what the correct behaviour is, here!

RAOF (changes requested):
Sorry, not approve! (Also not request changes, but that's the only way I can cancel my approval, I believe).

mattkae:
Abandoned this in favor of putting the fix in wayland-rs, as it appears that Gtk (at least) expects this event: https://matrix.to/#/!ivAdrrLoWkBYEnSIae:matrix.org/$br4dYeucp_lM4fe0JlN6Sh_uvuo4gptw9TJj2RKK3OY?via=matrix.org&via=ubuntu.com&via=fedora.im

PR #466, opened 2026-09-16
add workshop definition

andogq:
very similar to Mir's workshop

See canonical/mir#5237


saviq-work (changes requested):
Should we have a `compositor` mount point so you can mount the compositor build from another workshop and run that across workspaces without copying?

